PR #52, opened 2014-11-15
Fix broken HTML

nilicule:


kingosticks:
This fix seems to have already gotten in there. Thanks for highlighting it

Issue #63, opened 2015-01-05
Keyboard shortcut blocking


akkenoth:
Currently, MusicBox web interface blocks all keyboard shortcuts (like ctrl+tab, alt+w etc).

Expected behaviour would be only blocking these keyboard actions, that MusicBox overrides (ie space for pause/play).

Fix:
@mopidy_musicbox_webclient/static/js/gui.js:509
```
$(document).keypress( function (event) {
    //console.log('kp: '+event);
    if (event.target.tagName != 'INPUT') {
-       event.preventDefault();
        switch(event.which) {
            case 32:
                doPlay();
+               event.preventDefault();
                break;
            case '>':
                doNext();
+               event.preventDefault();
                break;
            case '<':
                doPrevious();
+               event.preventDefault();
                break;
        }
    return true;
    }
});
```

kingosticks:
This must be browser specific as those shortcuts work perfectly for me using Chrome Version 38.0.2125.111 (64-bit) on Linux Mint.

akkenoth:
As far as I understand the [specification](https://developer.mozilla.org/en-US/docs/Web/Events/keydown#preventDefault%28%29_of_keydown_event), preventDefault() should prevent all actions on particular event.

It is indeed browser specific, but in a reverse-ish manner: Chrome simply blocks preventDefault() on some keyboard events (see [this](http://stackoverflow.com/questions/7295508/javascript-capture-browser-shortcuts-ctrlt-n-w)).
I'm using Firefox (according to JS compliancy pages, version and OS are irrelevant), but other browsers like Safari/IE/etc can possibly have this issue too.

kingosticks:
Fair enough, just wanted to highlight the fact that verifying any fix would require a particular browser. 
If you can submit a pull request then it's nice and easy for @woutervanwijk

kingosticks:
Fixed in https://github.com/woutervanwijk/Mopidy-MusicBox-Webclient/commit/13c024464388fed85d060edbfe746926f1353a37

PR #64, opened 2015-01-09
Better handling of keyboard shortcuts and fixes removing tracks from queue

jcass77:
Explicitly translate key codes to actual characters before executing keyboard shortcuts. Current implementation seems to use a combination of key codes and characters in the switch statement which does not appear to work on all platforms (e.g. Safari on Mac OS).

Also contains fixes for issue #63, restores track removal functionality of issue #4, and addresses https://github.com/woutervanwijk/Pi-MusicBox/issues/118

### DIFF
--- a/mopidy_musicbox_webclient/static/js/controls.js
+++ b/mopidy_musicbox_webclient/static/js/controls.js
@@ -236,7 +236,7 @@ function removeTrack() {
     }
     var track = {};
     track.uri = [currentplaylist[i].uri];
-    mopidy.tracklist.remove(track);
+    mopidy.tracklist.remove({'uri':track.uri});
     //    console.log(currentplaylist[i].uri);
 }
 
--- a/mopidy_musicbox_webclient/static/js/gui.js
+++ b/mopidy_musicbox_webclient/static/js/gui.js
@@ -509,8 +509,10 @@ $(document).ready(function(event) {
     $(document).keypress( function (event) {
 	//console.log('kp:    '+event);
 	if (event.target.tagName != 'INPUT') {
-	    switch(event.which) {
-	        case 32:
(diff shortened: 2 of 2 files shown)


celeryclub:
These fixes work well for me. :+1:

mopidy_musicbox_webclient/static/js/controls.js
```diff
@@ -236,7 +236,7 @@ function removeTrack() {
     }
     var track = {};
     track.uri = [currentplaylist[i].uri];
-    mopidy.tracklist.remove(track);
+    mopidy.tracklist.remove({'uri':track.uri});
```
kingosticks: I'm no JS wizz but doesn't this work out exactly the same? The original code works just fine on my system.

kingosticks: Either way, in this case it's easier to just press the big merge button!

jcass77: It's interesting that the old code worked for you. Based on the [Mopidy API docs] ( https://docs.mopidy.com/en/latest/api/core/#mopidy.core.TracklistController.filter) I would expect 

```mopidy.tracklist.remove(uri = track.uri)``` 

to work as well (which is probably easier to read), but not just passing ```track```.

kingosticks: I'm presuming it works because javascript object notation is basically just
a dictionary and is seen as such by whatever magic js translation mopidy
does. But the latter won't.

jodal: It might have worked because of JS objects is simply "dicts", but the JSON deserializer on the server side is supposed to convert dicts with the `"__model__": "Track"` attribute back to a Python `Track` object before passing it to `remove()`.

Anyway, this is way more correct.

PR #66, opened 2015-01-25
Load js file through relative path

### LINKED ISSUE #65: Webclient broken if placed behind reverse proxy
In my setup, the mopidy server is made available through a reverse proxy (nginx) such that the mopidy root is available under `https://domain/mopidy`, i.e., the web clients are available as

 * `https://domain/mopidy/musicbox_webclient/`
 * `https://domain/mopidy/moped/`
 * ...

The webclient fails to start up because it looks for `https://domain/mopidy/mopidy.min.js` while it should load `https://domain/mopidy/mopidy/mopidy.min.js`.

The fix is to load the js file through the relative path `../mopidy/mopidy.min.js` instead of `/mopidy/mopidy.min.js`. I'll prepare a PR in a second.

andrenarchy:
Fixes #65

### DIFF
--- a/mopidy_musicbox_webclient/static/index.html
+++ b/mopidy_musicbox_webclient/static/index.html
@@ -457,7 +457,7 @@
 
 </div>
 <!-- /page one -->
-<script type="text/javascript" src="/mopidy/mopidy.min.js"></script>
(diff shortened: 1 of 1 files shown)


kingosticks:
Thanks!

kingosticks:
Forgot to actually merge it... Thanks again.

PR #86, opened 2015-03-17
Tweaks for better small screen layout and darker theme

southafricanrob:
1) volume slider movies to header on small screens for easy access
2) now playing moves to side panel on larger screens so always visible
3) added right side settings panel in header to free up some space from Now Playing panel 
4) dark theme option similar to Spotify Web Player
5) independently scrollable playlist and track list divs


kingosticks:
I think we will hold off on 1) and 2), for now at least. Sadly my attempts at cherry-picking the rest are meeting with little cooperation from git.

jcass77:
Closing this PR as these changes seem to have been taken further in #89 instead.

PR #89, opened 2015-03-24
Enhancement/small screen tweaks only

nickbarnwell:
I saw #86 was good minus the first couple of changes…this might successfully revert the commit. I couldn't set up a dev environment on this machine in the time I had, but figured you or Rob can probably quickly fix it up if I don't get to it before then.


nickbarnwell:
Hey @kingosticks, anything I can do make this easier for you to merge in? Would it help to rebase on top of the 2.0.0 tag?

kingosticks:
Hi, I do appreciate you taking the time to help with this. If you could rebase that'd be handy. We just did a release so I've steered clear of destabilising things but I'm keen to get this in now.

nickbarnwell:
Fantastic! Will try and take a crack at it tonight. Thanks for being such a responsive maintainer :+1:

camilonova:
@nickbarnwell any updates?

kingosticks:
On merging into develop now there's one conflict (the extra old code can be removed).

Having tried it again I've noticed some things:
* The header is now double height with a search icon in the middle. With the dark theme is looks like this is large white/grey button which fills the left half of the top header. Unsure of the intention here. 
* With the dark theme, the play button that lives on the left in the footer still has a blue background. Guess the image should be transparent.
* On my phone, the navigation doesn't work any more, only settings and system do anything. And the new settings icon on the header is not right aligned. With the dark theme, the settings icon is sitting inside a black rounded rectangle with a white border around it, perhaps it's a button.
* The scrollable divs are maybe just a bit too long and the bottom is cut off. So much so that you don't see the down arrow on the scrollbars. I think this is actually a long standing bug with the height of the inner content being slightly too big for the window so there's always a window scrollbar, even when there shouldn't be. It stands out more now due to the other extra scrollbars.
* The play and skip-forward icons on the footer are misaligned in the dark theme compared to the light theme. If you switch between them you can see them move to the right. It seems a bit weird that there's any positional information in the dark theme, is there a way to separate out the colours and the positional layout?

I believe both @woutervanwijk and myself liked the changes that we've got here, and while I appreciate these above issues are probably from the original PR, we can't be merging any broken UI code. So there's a bit more to do here - sorry.

PR #104, opened 2015-05-30
Add link to Alarm Clock

DavisNT:
Add link to [Mopidy-AlarmClock](https://github.com/DavisNT/mopidy-alarmclock) if it is enabled in Mopidy config (e.g. installed and loaded).

Related to https://github.com/woutervanwijk/Pi-MusicBox/issues/270

### DIFF
--- a/mopidy_musicbox_webclient/web.py
+++ b/mopidy_musicbox_webclient/web.py
@@ -45,7 +45,9 @@ class IndexHandler(tornado.web.RequestHandler):
             'version': MusicBoxExtension.version,
             'musicbox': int(ext_config['musicbox'] or False),
             'websocket_url': ws_url,
-            'alarmclock': int('alarmclock' in config and 'enabled' in config['alarmclock'] and config['alarmclock']['enabled'])
+            'alarmclock': int('alarmclock' in config and 
+                'enabled' in config['alarmclock'] and 
+                config['alarmclock']['enabled'])
         }
         self.__path = path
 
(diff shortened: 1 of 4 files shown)


mopidy_musicbox_webclient/web.py
```diff
@@ -41,10 +41,17 @@ def initialize(self, config, path):
                                'using %s', port)
             ws_url = "ws://%s:%d/mopidy/ws" % (host, port)
 
+        alarmclock = False
+        try:
+            alarmclock = config['alarmclock']['enabled']
```
kingosticks: A try block for this isn't really necessary.

DavisNT:
@kingosticks I removed try/except. I hope there won't be any cases when corrupted config will have non-dictionary types (and crash webclient).

kingosticks:
I think you can probably use get() with a default value, or whatever the
dictionary method is called. Mopidy sanitises the config anyway.

DavisNT:
If Mopidy always sanitizes the config (`config['alarmclock']` is always dictionary if present) then this code should work OK. I think that adding `get()` won't make code much shorter, but will add something that is not used in other places. Is current code OK to be merged?

kingosticks:
I think that's there now is nice and clear so that's fine by me.  

In the diff it looks like settings and shutdown now have slightly different ui-block-a and -b but it's hard to tell what's going on my phone and I can't remember the difference. But assuming that's nothing then I'll merge when I get home later.

DavisNT:
`ui-block-a` and `ui-block-b` determine alignment (left or right) if two columns are displayed. These classes are changed back in `gui.js` line 528 if Alarm Clock link is hidden.

kingosticks:
Oh yeh right those button things, thanks for the hint.

PR #108, opened 2015-06-14
Ability to save Queue as local Playlist

### LINKED ISSUE #106: Feature: Ability to save playlist
I would like to add ability to save queue as local playlist.

Queue could have a _Save_ button next to _Clear_ button. _Save_ button would open an input prompt with playlist name and save it as local playlist.

:question: Is such feature welcome? If yes, I will investigate it and, possibly, create a Pull Request.

DavisNT:
Fixes #106

### DIFF
--- a/mopidy_musicbox_webclient/static/js/controls.js
+++ b/mopidy_musicbox_webclient/static/js/controls.js
@@ -229,11 +229,12 @@ function clearQueue() {
 
 function saveQueue() {
     var plname = window.prompt("Playlist name:", "");
-    if (plname != null && plname != "") {
+    if (plname != null && plname.trim() != "") {
+        plname = plname.trim();
         mopidy.playlists.filter({"name": plname}).then(function(existing) {
             var exists = false;
             for (var i = 0; i < existing.length; i++) {
-                exists = exists || existing[i].uri.indexOf("m3u:")==0 || existing[i].uri.indexOf("local:")==0;
+                exists = exists || existing[i].uri.indexOf("m3u:") == 0 || existing[i].uri.indexOf("local:") == 0;
             }
             if (!exists || window.confirm("Overwrite existing playlist \"" + plname + "\"?")) {
                 mopidy.tracklist.getTracks().then(function(tracks) {
--- a/mopidy_musicbox_webclient/static/js/controls.js
+++ b/mopidy_musicbox_webclient/static/js/controls.js
@@ -227,6 +227,28 @@ function clearQueue() {
     return false;
 }
 
+function saveQueue() {
+    var plname = window.prompt("Playlist name:", "");
+    if (plname != null && plname != "") {
+        mopidy.playlists.filter({"name": plname}).then(function(existing) {
+            var exists = false;
+            for (var i = 0; i < existing.length; i++) {
+                exists = exists || existing[i].uri.indexOf("m3u:")==0 || existing[i].uri.indexOf("local:")==0;
+            }
+            if (!exists || window.confirm("Overwrite existing playlist \"" + plname + "\"?")) {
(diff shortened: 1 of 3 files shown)


mopidy_musicbox_webclient/static/js/controls.js
```diff
@@ -227,6 +227,28 @@ function clearQueue() {
     return false;
 }
 
+function saveQueue() {
+    var plname = window.prompt("Playlist name:", "");
+    if (plname != null && plname != "") {
+        mopidy.playlists.filter({"name": plname}).then(function(existing) {
```
kingosticks: Should we introduce new code that uses deprecated api methods?

kingosticks:
Also, eventually, we probably want this button to take us to our playlist editing page. But other that my previous comment, which can be dismissed if it's too much hassle, this looks great.

DavisNT:
I have just committed a fix for removal of whitespace characters surrounding entered playlist name.

1. Regarding `playlists.filter()` (filtering done on Mopidy side; API deprecated in Mopidy 1.0) it can be replaced by `playlists.getPlaylists()` (I am **not sure** whether this is also deprecated in Mopidy 1.0 or not), but this has negative effect on performance (because all playlists are sent from Mopidy to Webclient and filtering is done entirely on Webclient using JavaScript). This change is available for testing in another branch: https://github.com/DavisNT/Mopidy-MusicBox-Webclient/tree/feature/SavePlaylist_API In my case (with about 300 KB playlist data) using `playlists.getPlaylists()` it takes about one second until I receive prompt about existing playlist (with `playlists.filter()` I get the prompt instantly). I suggest keeping existing implementation as it works on both Mopidy 0.19 and 1.0 (AFAIK there are lots of deprecated APIs used elsewhere in Webclient), and does not have annoying delay.
2. Regarding playlist editing - I propose adding ability to delete local playlists and (at the beginning) do playlist editing in Queue. This would keep interface simple and intuitive (AFAIK similar to playlist editing in old versions of Winamp). Later we could add _Add to Playlist..._ option to songs in Queue and Playlists, and _Remove from Playlist_ option to songs in Playlists (AFAIK similar to interface of some leading audio players nowadays). I think that separate playlist editing page would add unnecessary complexity to UI.

kingosticks:
The Mopidy api docs specify what's deprecated. Using that will indeed kill
performance, 'as_list' is the new method. I know we don't use it elsewhere
yet, but adding new deprecated code makes more work later. Not a big deal,
maybe something to keep in mind next time. Thanks for your efforts, it's a
nice addition.

DavisNT:
Could you please merge this? :+1: 

`as_list` is introduced in Mopidy 1.0, so it would break compatibility with Mopidy 0.19. But enough about this :wink:

kingosticks:
0.19 comparability is not worth keeping ,but yes, this is fine and I'll
merge when I get to a pc.

DavisNT:
Thanks! :smile: :+1:

kingosticks:
Just did a final test before merging and I noticed that the behaviour when the queue is empty is probably wrong. I think the button should do nothing in this case, and maybe even be greyed out. I can take care of it, but what are your thoughts?

DavisNT:
I completely agree - this button should do nothing when Queue is empty. AFAIK greying it out would be inconsistent behavior, because other buttons (e.g. _Clear_ Queue button) are not grayed out when they are not applicable.

DavisNT:
Thanks for spotting the issue! Now this should be fixed. Can you please test and merge?

kingosticks:
Merged in https://github.com/woutervanwijk/Mopidy-MusicBox-Webclient/commit/ab59c7be007953de08f2f25287b0655c4bcc7ce0

PR #117, opened 2015-07-11
Feature refresh playlists

### LINKED ISSUE #107: Feature: Ability to refresh playlists
I would like to add ability to refresh existing playlists.

Playlists could have a _Refresh_ button in the top of screen (like queue has _Clear_ button). The _Refresh_ button would signal Mopidy to refresh existing playlists.

:question: Is such feature welcome? If yes, I will investigate it and, possibly, create a Pull Request.

DavisNT:
Fixes #107 

Please review and merge.


kingosticks:
Quick question: when I mouse-over on the refresh icon or click it, it changes colour and it looks like it's disabled (but it's not, it still works just fine). Is this just the hover and visited link css coming through?

DavisNT:
I think this is caused by `class="ui-link"` which is added to `
<a style="float:right; padding: 1.2em 10px" onclick="return refreshPlaylists();">` in DOM (most likely by jQuery Mobile).

PR #127, opened 2015-10-22
Fixed behavior of play now/next on empty tracklist

dmosen:
No track was added on empty playlist

### DIFF
--- a/mopidy_musicbox_webclient/static/js/controls.js
+++ b/mopidy_musicbox_webclient/static/js/controls.js
@@ -43,8 +43,6 @@ function playBrowsedTracks(action, trackIndex) {
             };
             mopidy.tracklist.index().then(function (currentTrack) {
                 mopidy.tracklist.add(null, currentTrack + 1, null, trackUris).then(maybePlay);
-            }, function() {
-                mopidy.tracklist.add(null, 0, null, trackUris).then(maybePlay);
             });
             break;
         case ADD_THIS_BOTTOM:
--- a/mopidy_musicbox_webclient/static/js/controls.js
+++ b/mopidy_musicbox_webclient/static/js/controls.js
@@ -36,13 +36,22 @@ function playBrowsedTracks(action, trackIndex) {
     switch (action) {
         case PLAY_NOW:
         case PLAY_NEXT:
-            mopidy.tracklist.index(songdata).then(function(currentIndex) {
-                var addFunc = mopidy.tracklist.add(null, currentIndex + 1, null, trackUris);
+            var playNowOrNext = function(index) {
+                var addFunc = mopidy.tracklist.add(null, index, null, trackUris);
                 if (action == PLAY_NOW) {
                     addFunc.then(function(tlTracks) {
                         mopidy.playback.play(tlTracks[0]);
                     });
                 };
+            };
+            mopidy.tracklist.getTracks().then(function(tracks) {
+                if (tracks.length === 0) {
+                    playNowOrNext(0);
+                } else {
+                    mopidy.tracklist.index(songdata).then(function(currentIndex) {
+                        playNext(currentIndex + 1)
+                    });
+                }
             });
             break;
         case ADD_THIS_BOTTOM:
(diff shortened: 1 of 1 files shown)


kingosticks:
Hi, thanks for this. I noticed this too the other day but hadn't had a look yet. I was thinking of changing the usage of the index method to not passing any arguments. This should then return the current songs index, or none if the tracklist is empty. The benefit being we don't have to transfer the entire tracklist which is relatively much slower. Any thoughts?

kingosticks:
Or change what you have here to a call to tracklist.getLength?

dmosen:
I absolutely agree. We can avoid the getTracks call. I wasn't really aware of the available mopidy.js functions. Implementing was though kind of trial and error.

After having a look into the documentation I found a (hopefully) better solution.

kingosticks:
I'm not too fond of using the error condition like that, it's akin to handling normal flow control with exceptions. Would you mind if we changed it to something like the following:
```
mopidy.tracklist.index().then(function (currentTrack) {
    if (!currentTrack) {
        mopidy.tracklist.add(null, 0, null, trackUris).then(maybePlay);
    } else {
        mopidy.tracklist.add(null, currentTrack + 1, null, trackUris).then(maybePlay);
    }
});
```

dmosen:
Hi, I know what you mean but I'm not aware of a better solution currently. The problem is that the `index` method calls the given "reject" function if it can't determine an index. Using the index function still seems to be the best opportunity to me.

Maybe I get something wrong but I suspect that your snippet wouldn't work as `currentTrack`'s value is always of type integer at that point of execution.

kingosticks:
Sorry I've been out of the country this last week. You are right, my snippet didn't work since `!currentTrack` is true when it's empty (great) but also when the first track is being played (not so great as the new track gets added before, rather than after, the current track). But if we go back to something like in my original comment that should work just fine since in the case of an empty tracklist we get `currentTrack == 0` and it doesn't matter what index you insert at, so using `1` is OK. e.g. 
```
mopidy.tracklist.index().then(function (currentTrack) {
    mopidy.tracklist.add(null, currentTrack + 1, null, trackUris).then(maybePlay);
});
```
When is the `index` function unable to "determine an index"?

dmosen:
Hi, I hope you enjoyed your trip. The snippet you posted equals what I implemented except the exception handling so this part is out of question from my side. ;-)

Regarding your question: Calling the `index` function without parameters gives you the index of the currently playing song. If you just started Mopidy no song is playing and thus no index can be given. The same may be true when you hit stop or clear the playlist (I checked this some time ago but can't remember exactly). In this "exceptional case" that the index is non-determinable I think the best we can do is trying to place the song at the beginning of the (empty or stopped) queue on "play next/now".

kingosticks:
> Calling the index function without parameters gives you the index of the currently playing song. 
Yes
> If you just started Mopidy no song is playing and thus no index can be given. 
Mopidy returns `None` which translates to `null` in javascript.
> The same may be true when you hit stop
No, in this case `index()` returns the index of the current track as normal.
> or clear the playlist.
Yes, in this case you also get `null`.
> In this "exceptional case" that the index is non-determinable I think the best we can do is trying to place the song at the beginning of the (empty or stopped) queue on "play next/now".
To be clear, no exception is thrown in this case and so the error(/reject?) code path doesn't execute. Since `null + 1 == 1` in JS, my above snippet does work. Have you tried it? If it doesn't work for you, can I ask what version of Mopidy you are running and what browser you are using?

dmosen:
Thanks for being so patient with me. I tested your proposal and it works fine for me too. Don't know when and why I ran into the exceptional state I mentioned above.

kingosticks:
No worries, I appreciate any help. I was concerned that maybe it was a
browser compatibility thing but fortunately not.

Thanks for being so patient with me. I tested your proposal and it works
fine for me too. Don't know when and why I ran into the exceptional state I
mentioned above.

—
Reply to this email directly or view it on GitHub
<https://github.com/pimusicbox/mopidy-musicbox-webclient/pull/127#issuecomment-155147045>
.

kingosticks:
Merged in https://github.com/pimusicbox/mopidy-musicbox-webclient/commit/aa7b0a59da3159aa610c108980c095fc452f5294

PR #137, opened 2016-01-24
General cleanup and small tweaks

jcass77:
I noticed that there are a bunch of artefacts in the repository that are no longer used (old stylesheets, unused javascript, etc.).

This PR cleans out a bunch of them which reduces some clutter in the repository and should make it easier for new contributors to get up to speed with what's going on.

I also made a few small changes and applied fixes while I was working through the code:

- usage of ``mb.manifest`` has been re-instated. This should allow web caching to work as expected. But more importantly, it also ensures that IOS clients that have added the web client to their home screen will be able to have their caches refreshed in order to see any changes that are released as part of future versions. (see https://gregsramblings.com/2012/05/28/html5-application-cache-how-to/).

- increases the slider handle size by 30%, which makes it a little easier to grab on mobile devices. I think it is still small enough not to look out of place in the UI but we can tweak this further if anyone feels differently.

- use 'min' versions of the stylesheets throughout.

- fix alignment of playback control buttons in footer. The control buttons on the right (i.e. repeat, shuffle, etc) were being rendered noticeable higher than the playback buttons on the left. Also, the```-webkit-baseline-middle``` alignment that was used before is not supported on most browsers as far as I can tell.

- fixed the tag nesting of the form contained in https://github.com/pimusicbox/mopidy-musicbox-webclient/blob/0b94bd4071dee4a2f788ca6e1b4963950ce12384/mopidy_musicbox_webclient/static/index.html#L414-L433

Comparison of the old and new layout is attached.

- Old: 
img_1152

- New: 
img_1149


jcass77:
Note: don't think the travis build errors are due to anything I changed. Might be related to the mopidy packages that are used in the build?
```
The following packages have unmet dependencies:
 mopidy : Depends: python:any (>= 2.7.5-5~)
          Recommends: gstreamer0.10-alsa but it is not going to be installed
          Recommends: gstreamer0.10-plugins-ugly but it is not going to be installed
          Recommends: gstreamer0.10-pulseaudio but it is not going to be installed
          Recommends: gstreamer0.10-tools but it is not going to be installed
E: Unable to correct problems, you have held broken packages.
The command "sudo apt-get install mopidy" failed and exited with 100 during .
Your build has been stopped.
```

jcass77:
I pushed one more commit with the following changes:

- ensure that the height of the 'normalFooter' and 'nowPlayingFooter' is always aligned.
- remove the pause/play images that were still being used in the 'normal' footer in favour of standardising everything on the font-awesome icons.
- remove a buch of unused stylesheet elements.

There's still **a lot** of cleanup that can be done on this extension. But I'll stop here to give everyone a chance to review and comment before fiddling with anything else.

kingosticks:
Looks like Travis is happy again.

This is all perfect, thanks so much.

PR #139, opened 2016-01-24
Fix mute/unmute feature

jcass77:
The mute/unmute feature doesn't work as expected anymore. This PR applies the following fixes:

- swops the mute/unmute icons so that ``fa-volume-off`` is used to mute and ``fa-volume-up`` is used to unmute. This seems more intuitive.

- Unmuting now correctly restores volume to what it was before mute was toggled.

- User can now also slide volume up to do a manual unmute, and the mute/unmute functionality will still work as expected from there on out.

I was a little overzealous and removed two 'play' icons in my previous PR that is used to highlight the currently playing track in the queue. I've added them back again as part of this PR.


kingosticks:
On my system there is no mute icon unless I mouse-over the area, at which point the icon becomes visible. I'm not sure if it was like this before, I can't say I've used the mute function in years.

jcass77:
Yes I believe it has always been like that. The mute icon is not hidden explicitly but different browsers are interpreting it differently. It is visible in Safari, but hidden in Chrome, Firefox, and Safari IOS for example.

I think this has got something to do with the icon's colour and how the theme is set up, but haven't looked into it.

I've pushed another commit that at least gives it a proper label, and also refactors some of the volume slider handling code to reduce the reliance on global variables.

jcass77:
There are probably a few other mouse-over issues elsewhere as well. I noticed that some of the icons have different colours, or do not change colour on mouse-over, even though they appear to be defined identically in the HTML code and stylesheets?

The playback controller buttons in the footer are good examples (taken from Chrome v 47.0.2526.111):

skitched-20160125-080525

I do not have the faintest idea where to start looking for this. The jquery libraries probably haven't been updated in a while - not sure if that might solve the issue.

jcass77:
Disregard my last comment: it seems that the ```setTracklistOption()``` javascript function is used to set the button colors. Everything that cannot be toggled permanently ends up rendering as a slightly lighter shade of blue.

So this is normal but 'broken' in Safari.

jcass77:
I think this PR is ready for final review and merge. In the end I just reverted to calling ```mopidy.mixer.setMute()``` which I believe is the more standard approach to take rather than fiddling with volume levels directly.

jcass77:
...last minute change: I swopped the mute / unmute icons back to the way they were before to correspond to each state.

I don't know why I thought this would be more intuitive - it obviously isn't.

kingosticks:
I think we need something here to check the mute value on page load and respond to `muteChanged` events. If you set mute, and then open a new browser window the icon is wrong. Similarly, with two browser windows open and enabling mute in one of them. the mute icon in the other doesn't change.

jcass77:
Ok, I've opened #141 so that we can track and resolve that separately.

Thank you.

PR #140, opened 2016-01-26
Split and update libraries

jcass77:
This is the second part of a general cleanup of the Mopidy-Musicbox-Webclient codebase.

- all of the various libraries have been split into their constituent parts in the ``vendors`` folder. This should hopefully make it easier to identify and maintain them going forward.

- all of the libraries have been upgraded to the latest versions. This generated the bulk of the diffs which are like for like replacements of the old versions without exception.

- note that jQuery-Mobile itself cannot be upgraded since the Mobile Flat-UI Theme only supports version 1.3.x. (see https://github.com/ququplay/jquery-mobile-flat-ui-theme#current-branch-layout). It seems unlikely that the theme will be updated anytime soon so we may ultimately have to switch to something that is being maintained more actively if we want to carry on with jQuery-Mobile.

- some files were reformatted where it seemed appropriate, and old comments removed.

- applied a few small fixes to scripts (mainly typos or invalid arguments).

Once this is in we should be able to go back to just generating small incremental diffs as we work towards closing some of the open issues.

I did not encounter any issues post the upgrades, but it may be a good idea if a few more people could try it out for a day or two before we merge?


jcass77:
@kingosticks, I've rebased this on the latest develop branch. Once the 2.1.0 release is out, perhaps take a quick look at this to see if you are comfortable to merge?

This should make it much easier to update stylesheets and look at upgrading or replacing individual vendor-provided libraries, so I am keen to get it in before starting on any other issues...

jcass77:
I've rolled back the jquery-cookie libraries to v1.4.1. For the life of me I couldn't get the streamUri upgrade stuff (#152) to work with the latest version [2.1.0](https://github.com/js-cookie/js-cookie)

jcass77:
Now that 2.1.0 is out I think it is safe to move this over to develop.

Issue #141, opened 2016-01-26
Mute icon state not set correctly between browser windows


jcass77:
If you set mute, and then open a new browser window the icon is wrong. Similarly, with two browser windows open and enabling mute in one of them, the mute icon in the other doesn't change.

We probably need to check the mute value on page load and respond to ```muteChanged``` events.

PR #142, opened 2016-01-27
Add event handling for 'muteChanged' event

jcass77:
Proposed fix for #141 - handles ```muteChanged``` events.


jcass77:
I'd like to propose that we do a 2.0.1 bugfix release when this has been merged, and then migrate to the new libraries and clean up a few high priority issues in a proposed 2.1.0 release.

@kingosticks, please take a look at the milestone assignments to see if you agree?

mopidy_musicbox_webclient/static/js/controls.js
```diff
@@ -424,15 +424,19 @@ function triggerVolume() {
     volumeChanging = 0;
 }
 
-function toggleMute() {
-    mopidy.mixer.getMute().then(function(mute) {
-        mopidy.mixer.setMute(!mute);
+function setMute(nwmute) {
```
kingosticks: Nitpicking really but it would be clearer to call this something like `updateMuteIcon` and distinguish it from `doMute`.  I think there are some things that follow a convention like this, but I might be remembering that wrong. It would be nice to have one.

jcass77: It took me a while to figure out initially, but that appears to be the convention (which we are following here as well):

- ```doFoo``` is used in the HTML to trigger the actual Mopidy API calls.

- ```setFoo``` is only used to update the UI and change icons or toggle colours etc.

I agree that it would be clearer if we could change this, and maybe that is something that we can refactor across the entire codebase as part of the general bug cleanup. 

It's beyond the scope of this PR though.

kingosticks:
Github went down as i was commenting last night. The only other thing I was gonna say is that let's just make it the 2.1.0 release since there's more than bug fixes present.

jcass77:
@kingosticks, do you mean that we should release a 2.1.0 version now and then plan 2.2.0 for the new libraries and other bug fixes?

We just need to be sure that this PR is included in whatever the next release is as it contains fixes to #139 which has already been merged (sorry I was a little impatient).

kingosticks:
Yes that's what I mean.

kingosticks:
And yes, this will go in 2.1.0. I'd just like to test it out and make sure everything (I can think of) is now sorted with regards to this. I should get a chance to do so in the PM.

jcass77:
Ok, I'll update the milestone labels so long and see if there is anything else that should be documented in the 2.1.0 changelog.

mopidy_musicbox_webclient/static/js/controls.js
kingosticks: Yep, oops, you are of course totally right. We should look at making it more obvious at some point in the future.

mopidy_musicbox_webclient/static/js/controls.js
```diff
@@ -424,15 +424,19 @@ function triggerVolume() {
...
             $("#mutebt").attr('class', 'fa fa-volume-off');
+        } else {
+            $("#mutebt").attr('class', 'fa fa-volume-up');
         }
-    });
+    }
+}
+
+function doMute() {
+    mopidy.mixer.setMute(!mute).then();
```
kingosticks: Don't think we need to worry about the promise here

jcass77: Yes I saw this used elsewhere and wasn't sure - will you just remove after the merge?

PR #145, opened 2016-01-28
Increase available width of header.

jcass77:
This PR removes the ``ui-grid-b`` attribute which divides the screen into columns depending on how many child HTML elements are being rendered.

Since we are rendering two links and one header element, only 33.33% of the screen width is available for the title which seems unnecessary.

Also fixed reference to minimised version of ``jquery.mobile.iscrollview.css`` which doesn't exist. This begs the question if this stylesheet is even relevant at this stage? 

I think we can still sneak this PR in for the 2.1.0 release, as it is so small.


kingosticks:
I have a vague recollection of @woutervanwijk mentioning ages ago that the iscrollview stuff was removed after some issues with it. Any Ideas, Wouter?

jcass77:
I'll remove ```iscrollview``` from #140 when I rebase it after the release of 2.1.0.

If I understand it correctly, then it could be used to do things like 'pull-down-to-refresh' on mobile devices, which would certainly be handy. But we can rather start afresh to see if if is doable (if not with ```iscrollview``` then perhaps something else).

jcass77:
...this probably implies that ```iscroll-lite.js``` isn't really relevant at this stage either.

woutervanwijk:
Yes I removed it because it did more harm than good. I used it to have two panels which could separately  scroll on larger devices, so that the user could scroll the list of playlists independently from the tracks of the playlist. It's more user-friendly. But I had trouble implementing it well on phones and pc/tablets at the same time.

kingosticks:
That independent scrolling is something I *really* want to bring back. There was a PR from someone but it had some problems when I tested it. Unfortunately I can't remember exactly what.

PR #150, opened 2016-01-30
Implement the new 'by-position-or-by-name' Mopidy.js calling convention

### LINKED ISSUE #79: Use explicit Mopidy.js calling convention
In Mopidy.js 0.4 / Mopidy 0.19 a new "calling convention" setting was added to Mopidy.js. See https://docs.mopidy.com/en/latest/api/js/#creating-an-instance for details.

It seems MusicBox-Webclient still doesn't explicitly set the calling convention of Mopidy.js, so it defaults to "by-position-only". This default will change in the future, and when that happens, MusicBox-Webclient will break as is.

In the Mopidy.js docs, there's some motivation for changing to the new alternative calling convention, "by-position-or-by-name". The migration to it can be as simple as adding `[]` brackets around all existing Mopidy.js calls.

jcass77:
Fixes #79.

I also did a bit of reformatting and cleaned up some old comments while I was working through the code. But from a functional point of view the calling convention is the only thing that changed.


jodal:
Great to see MusicBox Webclient move to the new calling convention. One less thing that'll break when we remove deprecated arguments :)

jcass77:
I've pushed a81a3ac1d84205e1ee90eff7d7fdd07879a42601 which replaces the calls containing the deprecated ``uri`` parameter to ``core.LibraryController.lookup()``.

Seems to be working ok, I'm going to merge this in the next while if there are no objections?

mopidy_musicbox_webclient/static/js/library.js
```diff
@@ -263,7 +266,8 @@ function showArtist(nwuri) {
 //TODO cache
     $('#h_artistname').html('');
     showLoading(true);
-    mopidy.library.lookup(nwuri).then(function(resultArr) {
+    mopidy.library.lookup({'uris': [nwuri]}).then(function(resultDict) {
+        var resultArr = resultDict[Object.keys(resultDict)[0]];
```
kingosticks: Any reason you didn't just go for `var resultArr = resultDict[nwuri]` ?

jcass77: I got tripped up on how the callbacks were handled in https://github.com/pimusicbox/mopidy-musicbox-webclient/blob/4992d038a950d69a00077ea7e5e80d7b31b72dfa/mopidy_musicbox_webclient/static/js/process_ws.js#L121-L125 and figured that I would just get the URI from the result dictionary every time to be safe.

You're right that it is not necessary here (and in a few other places) and is easier to read. So I've changed it in 4ec547a66ac24561fb1e58f9f9b4f76a031ccf08.

PR #152, opened 2016-02-01
Upgrade streamuris automatically

kingosticks:
A quick solution to automatically convert a users existing streamuris (stored in their browser as a cookie) to a local m3u playlist. This, as well as the new favourites functionality in general, might be broken if a user has the m3u extension disabled....


mopidy_musicbox_webclient/static/mb.manifest
```diff
@@ -1,6 +1,6 @@
 CACHE MANIFEST
 
-# 2016-01-31:v3
+# 2016-02-01:v1
 
 NETWORK:
 *
```
jcass77: We probably want to remove ``js/streamuris.js`` from the manifest file as well.

``printf "CACHE MANIFEST\n\n# 2016-MM-DD:vN\n\nNETWORK:\n*\n\nCACHE:\n" > mb.manifest; find . -type f | sed "s#^\./##" | grep -vi "ds_store" >> mb.manifest`` can also be run from the ``static`` directory to regenerate the entire manifest file in situations where that would be easier than updating it by hand.

mopidy_musicbox_webclient/static/js/controls.js
```diff
@@ -604,6 +613,36 @@ function showFavourites() {
     });    
 }
 
+function upgradeStreamUrisToFavourites() {
+    $.cookie.json = true;
```
jcass77: Is this function something that we should flag for removal with a TODO type statement? I assume once everyone has transitioned to >= v2.1.0 this won't be needed anymore?

Maybe in release 3.x or 4.x?

kingosticks: Yes, good idea.

mopidy_musicbox_webclient/static/mb.manifest
kingosticks: I'm all for automating this somehow.

mopidy_musicbox_webclient/static/js/controls.js
```diff
@@ -604,6 +613,36 @@ function showFavourites() {
...
+                uris.push(rs[1]);
+            }
+        }
+        mopidy.library.lookup({'uris': uris}).then(function(results) {
+            var tracks = []; // Prepare a list of tracks to add.
+            for (var key in streamUris) {
+                var rs = streamUris[key];
+                if (rs) {
+                    var track = results[rs[1]][0];
+                    track.name = rs[0] || track.name; // Use custom name if provided.
```
jcass77: This fails for me if ```mopidy.library.lookup()``` could not retrieve a track for the given URI. And the exception causes the entire loop to abort and not do ```addToFavourites(tracks);``` for any of the streams in the cookie (so nothing is added to the favourites playlist).

The easiest way to replicate this is to save a geo-blocked youtube stream to the cookie and then try to upgrade (in my case that is one of the standard ones distributed in ```streamuris.js```: http://www.youtube.com/watch?v=eQsM6u0a038

jcass77: ...might be as simple as adding the following?
```javascript
 if (track) {
    track.name = rs[0] || track.name; // Use custom name if provided.
    tracks.push(track);
    } else {
        console.log("Skipping unplayable streamUri " + rs[1]);
    }
```

mopidy_musicbox_webclient/static/mb.manifest
jcass77: I had a brief look around but could not find anything on an *easy* way to generate the manifest file automatically (without resorting to installing weird and wonderful add-ons and plugins).

Unless anybody has any ideas, I suggest we just keep this a manual step for now...

kingosticks: Despite the noise it generates I'm not against trying to keep doing this, I can just see myself forgetting.  I guess all we *really* need to ensure is that the releases that we package / tag get the treatment.

Or maybe Travis could get off his arse and check the manifest as part of his build :)

mopidy_musicbox_webclient/static/js/controls.js
kingosticks: Good catch.

But doesn't the exception in `lookup` still skip the addToFavourites call with this?

kingosticks: I guess it also makes more sense to invert this and iterate over the lookup results instead.

jcass77: No, ``lookup`` doesn't raise an exception as I originally thought, it just returns a dictionary without a value for the URI key. 

So this works as far as I can tell.

jcass77: ...in other words, the exception is raised in the JavaScript.

kingosticks:
Hopefully this now handles when the m3u extension is disabled.

mopidy_musicbox_webclient/static/js/controls.js
```diff
@@ -588,22 +607,65 @@ function deleteFavourite(index) {
...
+                child += '<h1>' + favourites.tracks[i].name + '</h1></a></li>';
                 tmp += child;
             }
-            $('#streamuristable').html(tmp);
         }
+        $('#streamuristable').html(tmp);
     });    
 }
 
+# TODO: Remove this upgrade path in next major release.
```
jcass77: This produces a syntax error. Should probably be:
```javascript
/* TODO: Remove this upgrade path in next major release. */
```
Apart from that I think we should merge.

jcass77: ...and maybe just add a not to the changelog mentioning that users should click on the **Convert streamUris** button after they have upgraded?

kingosticks: The fact I was trying to write a different language explains a lot.

kingosticks:
Unsure what githubs merge conflict problem is

jcass77:
Luckily, I think it is just the version number on ``mb.manifest``.

kingosticks:
That's pretty annoying, another reason to try and find a way around updating this manually all the time.

PR #153, opened 2016-02-04
Bump Mopidy version dependency for M3U-related features.

### LINKED ISSUE #147: Latest webclient requires a newer version of Mopidy than v0.19
We currently require at least Mopidy v0.19 but I think with some of the more recent changes that needs to be increased to at least v1.0.

jcass77:
Fixes #147.


jcass77:
Merging this as it has already been discussed in #147.

Issue #53, opened 2014-11-17
Problem with fullscreen mode


theactiveone:
Hey, 

I think there is a bug in the online user interface. In fullscreen mode I can't scrool, even if the list of tracks or something like this is longer as the viewpoints height.

I hope u wil fix it.

jcass77:
Same problem under Safari on OS X.

PR #158, opened 2016-02-05
Prevent mobile devices from scaling up when in landscape mode.

### LINKED ISSUE #157: Fullscreen not scaling correctly in landscape mode
When running the web client in fullscreen mode on a mobile device, switching to landscape mode sometimes causes the page to scale incorrectly.

I've seen this on the iPhone 5s but assume the same issue will occur on other mobile devices with similar screen dimensions:

img_1169

Steps to reproduce:

1. first add Mopidy-Musibox-Webclient to the Home Screen, and then launch the client.
2. click on the search icon in the top right corner of the screen.
3. make sure that the 'Search term' input box has focus (i.e. the keyboard is displayed, ready for text entry).
4. click 'done' to make the keyboard go away.
5. rotate device to enter landscape mode.
6. incorrectly scaled screen is displayed as per the above image.

What makes this particularly annoying is that, once the scaling has gone haywire, there is no way to recover in either landscape or portrait mode. The client needs to be restarted to reset to the correct viewport.

jcass77:
This avoids the landscape zoom-in effect on the iPhone. Perhaps somebody with a different device can give it a quick try to see if produces the same result?

Fixes #157.


jcass77:
Any ideas why git is flagging all of my commits as being @wagamama's?

jcass77:
I'm probably going to merge this in the next day or two if there are no objections...

kingosticks:
I can't test this right now at work but I'll try to do so on a couple of small devices tonight. 

As for the weird commit collaboration - no idea.

kingosticks:
I had a play on my Android phone (v5.1, chrome) and I don't see the original problem, but I see no problems with these changes either so go for it. However, I did find that I'm completely unable to scroll when in fullscreen mode on my Android phone which makes it pretty unusable. I never noticed this issue before but I've never tried to use fullscreen mode on my phone (it's arguably a strange thing to do since the way the latest apps works, it's just not necessary. It's out of scope for this PR either way.

jcass77:
Sounds like you may be experiencing #53 on mobile as well.

PR #159, opened 2016-02-07
Avoid polling for current track and time changes.

### LINKED ISSUE #40: Avoid polling for current track and time changes when not playing.
Mopidy will let you know over the WS when the state changes, so you should be able to get rid of the polling. Ideally you check this type of state once the connection is up to get an initial state, and then after that you just let us emit events to you when things change.

Only exception to this might be that you still want to fetch the time every now and then while playing to avoid drifting, as won't haven't figured out a better solution for this quite yet.

jcass77:
Based on the work done by @wagamama and @adamcik.

- Replaces existing track slider timer with @adamcik's [media-progress-timer](https://github.com/adamcik/media-progress-timer).

- Checks to ensure that the timer is in sync with ``mopidy.playback.getTimePosition()``. This is useful for streams that take a long time to buffer. Ideally, Mopidy should have an audio listener that triggers an event when the buffer has been filled and gstreamer actually starts to play the track?

- Removes reliance on ``updateStatusTimer`` for polling Mopidy to check the status of the currently playing track and playback position.

Fixes #40, replaces #95.

### DIFF
--- a/mopidy_musicbox_webclient/static/js/controls.js
+++ b/mopidy_musicbox_webclient/static/js/controls.js
@@ -260,11 +260,14 @@ function setPlayState(nwplay) {
         $("#btplayNowPlaying").attr('title', 'Pause');
         $("#btplay >i").removeClass('fa-play').addClass('fa-pause');
         $("#btplay").attr('title', 'Pause');
+        mopidy.playback.getTimePosition().then(processCurrentposition, console.error);
+        startProgressTimer();
     } else {
         $("#btplayNowPlaying >i").removeClass('fa-pause').addClass('fa-play');
         $("#btplayNowPlaying").attr('title', 'Play');
         $("#btplay >i").removeClass('fa-pause').addClass('fa-play');
         $("#btplay").attr('title', 'Play');
+        progressTimer.stop();
     }
     play = nwplay;
 }
@@ -357,40 +360,13 @@ function doSingle() {
  * Use a timer to prevent looping of commands  *
  ***********************************************/
 function doSeekPos(value) {
-    var val = $("#trackslider").val();
-    newposition = Math.round(val);
-    if (!initgui) {
-        pausePosTimer();
-        //set timer to not trigger it too much
-        clearTimeout(seekTimer);
-        $("#songelapsed").html(timeFromSeconds(val / 1000));
-        seekTimer = setTimeout(triggerPos, 500);
-    }
-}
-
-function triggerPos() {
     if (mopidy) {
-        posChanging = true;
-        mopidy.playback.seek({'time_position': newposition});
-        resumePosTimer();
-        posChanging = false;
+        mopidy.playback.seek({'time_position': Math.round(value)});
     }
 }
 
 function setPosition(pos) {
-    if (posChanging) {
-        return;
-    }
-    var oldval = initgui;
-    if (pos > songlength) {
-        pos = songlength;
-        pausePosTimer();
-    }
-    currentposition = pos;
-    initgui = true;
-    $("#trackslider").val(currentposition).slider('refresh');
-    initgui = oldval;
-    $("#songelapsed").html(timeFromSeconds(currentposition / 1000));
+    setProgressTimer(pos);
 }
 
 /***********************************************
@@ -428,33 +404,6 @@ function setMute(nwmute) {
 
 function doMute() {
     mopidy.mixer.setMute({'mute': !mute});
-}
-
-/**************************
- *  Track position timer  *
- **************************/
-
-//timer function to update interface
-function updatePosTimer() {
-    currentposition += TRACK_TIMER;
-    setPosition(currentposition);
-}
-
-function resumePosTimer() {
-    pausePosTimer();
-    if (songlength > 0) {
-        posTimer = setInterval(updatePosTimer, TRACK_TIMER);
-    }
-}
-
-function initPosTimer() {
-    pausePosTimer();
-    // setPosition(0);
-    resumePosTimer();
-}
-
-function pausePosTimer() {
-    clearInterval(posTimer);
 }
 
 /************
--- a/README.rst
+++ b/README.rst
@@ -69,6 +69,7 @@ v2.2.0 (UNRELEASED)
 - Remove unused iScroll libraries and references.
 - Remove unused jQuery.Mobile.iScrollView libraries and references.
 - Remove unused jQuery.Truncate libraries and references.
+- Avoid polling for current track and time changes. (Fixes: `#40 <https://github.com/pimusicbox/mopidy-musicbox-webclient/issues/40>`_).
 
 **Fixes**
 
--- a/mopidy_musicbox_webclient/static/css/webclient.css
+++ b/mopidy_musicbox_webclient/static/css/webclient.css
@@ -179,11 +179,6 @@
     display: none !important;
 }
 
-#songelapsed, #songlength {
-    font-size: 10px;
-    margin-top: 12px;
-}
-
 /************************
  *  Volume Slider
  ***********************/
@@ -432,10 +427,14 @@ a {
 
 .pull-right {
     float: right;
+    font-size: 10px;
+    margin-top: 12px;
 }
 
 .pull-left {
     float: left;
+    font-size: 10px;
+    margin-top: 12px;
 }
 
 .hidden, #allresultloader, .loader {
--- a/mopidy_musicbox_webclient/static/index.html
+++ b/mopidy_musicbox_webclient/static/index.html
@@ -297,10 +297,10 @@
     </div>
 
     <div id="slidercontainer"><!-- slider for track position -->
-        <span id="songelapsed" class="pull-left"> 0:00 </span>
-        <span id="songlength" class="pull-right"> 0:00 </span>
+        <span id="songelapsed" class="pull-left"></span>
+        <span id="songlength" class="pull-right"></span>
         <label for="trackslider" disabled="disabled" class="ui-hidden-accessible">Position</label>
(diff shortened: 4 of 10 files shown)


kingosticks:
Nice one. I'll test this ASAP.

jcass77:
Just a note that this uses a patched version of media-progress-timer (see: [media-progress-timer/pull/2](https://github.com/adamcik/media-progress-timer/pull/2)).

I'd be interested to know if anybody has any ideas of how this can be handled better / differently.

mopidy_musicbox_webclient/static/js/controls.js
```diff
@@ -357,40 +357,18 @@ function doSingle() {
...
-    var oldval = initgui;
-    if (pos > songlength) {
-        pos = songlength;
-        pausePosTimer();
+    synced = false;
+    progressTimer.set(pos);
+    if (!play) {
+        // Set slider position directly as timer callback is not currently
+        // running.
+        $("#trackslider").val(pos).slider('refresh');
```
adamcik: `timer.set(...)` runs the callback, so you shouldn't need to does this manually.

jcass77: I think this relates to the comments that I made on the fact that the timer has been patched.

The timer can't always be used to update the slider as it interferes with user events. 

But I could also just be being dense...

adamcik:
Is https://jsfiddle.net/00sg7yff/ comparable to the problem you see in using the timer? If so I would try something like https://jsfiddle.net/qgp2zpts/

Basic idea is that when you start dragging the slider or changing it in any way you call `stop()`. As the position changes you call `set(position)` updating the UI via the timer. Then when you are done sliding things around you trigger a seek against mopidy but leave to timer stopped. Then when you get the `position-changed` event you call `timer.set(position).start()`

This is at least how I've gone about in an other project which uses the media-progress-timer. The possible shortfall in this is that you don't get the playing state and the position in the same event, which we need to get fixed in mopidy. But for now you could probably work around that.

As the buffering bit, that also plays in to the same type of event overhaul which we need.

adamcik:
If that makes somewhat sense I'll try and add it to the docs for the lib as this is probably a stumbling block for anyone else trying to use it :-)

jcass77:
@adamcik, https://github.com/jcass77/Mopidy-MusicBox-Webclient/blob/182d4822a27d8cba9480c646f494d8b711a0e932/mopidy_musicbox_webclient/static/js/gui.js#L599-L605 handles the slider event bits as you described (except that it also uses and internal state to resume the timer if Mopidy is playing).

Isn't ``seeked`` the event to listen for instead of ``position_changed``? I thought that was was is used for manual seeks instead of just position updates?

The issue with all of this is still that https://github.com/adamcik/media-progress-timer/blob/28007e9c86d09cc556a9a47f6c6c9fc39cc18c2d/timer.js#L71-L75 calls the callback again, even though the timer was stopped in the ``slidestop`` listener. So fi the user is using the callback to update the slider position in the UI, it will override whatever the user is trying to do while dragging the slider around.

Make sense?

adamcik:
Indeed, I'm thinking in terms of the audio subsystem names, in core it is indeed 'seeked'. I've just been to deep into audio lately.

Ah, I see. So in theory doing a `function stop(update_position)` might also workaround the issue then. I need to play around with my test code and other places I use this a bit more to double check what makes sense. I'll probably move the rest of the followup to the other PR you opened. Just started here as it's a bit easier when seen in context.

adamcik:
Oh and as for the `if (playing) { timer.start() }`, I would recommend postponing that until the `seeked` event. Otherwise you will be ahead by whatever amount of time it takes you to perform the seek. Or you would jump back again if the `seeked` event sets the position.

adamcik:
Made https://jsfiddle.net/0qqypp55/ which tries to emulate the musicbox code more closely. Having added the 100ms timeout for the seek to happen I now have it glitching back and forth once. Is this what you are trying to avoid?

jcass77:
I couldn't pick up the glitch at 100ms but setting it to 1000ms: yes, that is more or less the animation effect that I am trying to avoid. On a conceptual level, what I am aiming for is (assuming that the user is playing a stream that takes 5-10 seconds to buffer on every seek):

1. Mopidy playback starts and triggers ```playbackStateChanged```, which starts the timer.
2. Instead of progressing the slider immediately, the timer callback starts doing ```mopidy.playback.getTimePosition()``` until the position moves off zero.
3. The callback now starts progressing the slider, and should in theory be in sync with Mopidy's position.

Doing a seek by dragging the slider should work in a similar way as the above:

1. User drags the slider, and  ``ui.value`` (or ``$(this).val()``?) should contain the target position value.
2. ``mopidy.playback.seek(ui.value)`` is called to perform the seek.
3. The timer does the callback and waits for ```mopidy.playback.getTimePosition()``` to match whatever ``ui.value`` returned before progressing the slider.

In practice it doesn't appear to be working at all unfortunately. The results that I am getting with ``mopidy.playback.getTimePosition()`` seem a little unpredictable between track changes (probably because I am not using the promise correctly). So the slider always moves off zero and then glitches back once the track audio actually starts.

jcass77:
...what would be really great for consumers of media-progress-timer is if they could pass it an optional reference timer that can be used as a heartbeat to keep it in sync with the actual source.

In this case it would call ```mopidy.playback.getTimePosition``` to start off in sync, and then again every few seconds or n-number of callback cycles to check that is is still in sync, or whenever the source indicates that a state change has occurred that requires a resync.

This is basically what I am trying to accomplish in the Muiscbox-Webclient. And failing is spectacular fashion.

jcass77:
I think my completely-over-engineered-track-progress-timer is ready for review with the push of https://github.com/jcass77/Mopidy-MusicBox-Webclient/commit/67706a2202a4d1c77dc4c7725a86fc45a746f5d7.

Changes:

- now includes a few wrapper functions for media-progress-timer to implement the 'reference timer / heartbeat' idea. I'm sure these could be integrated in a more useful form if there is sufficient interest for people to re-use.

- the progress timer now accounts for any buffering that Mopidy may be doing in order to give the **actual** time position, which should hopefully be within 500ms of the audio.

- it includes an exponential step-back so that the sync is checked less and less often as the track progresses and the progress timer remains in sync. The default is: 500ms, 1, 2, 4, and 8 seconds and then every 30 seconds thereafter while the track is playing. This is just a workaround while the Mopidy events described in https://github.com/mopidy/mopidy/issues/1420 are not available.

- I used the time elapsed indicator on the left of the slider to output some debug info while doing the development, and in the end I thought these were useful enough (and kinda cool looking) to consider keeping in the final product:

  - ``(wait)`` - progress timer is waiting for Mopidy to provide the target position.

  - ``(sync)`` - timer is busy syncing (i.e. Mopidy is buffering).

  - ``m:ss`` - timer is in sync and elapsed time is displayed normally.

kingosticks:
Works great on my desktop from playing around for about 10 minutes. 

When I try on my phone I found that if pause the music and then tap quickly a few times at different points on the slider, it'll start moving again as if I unpaused the music (which I didn't). It moves for a certain amount of time and then flicks back to the last placed I tapped, this process then repeats. I presume that when it erroneously  starts moving it requests a position update and then when it receives this, this causes it to 'flick' back. And this just continues over and over since the slider moving state is out of sync with Mopidy's play state.

I am not able to reproduce this on my desktop, but I can't really tap in many places quickly in the same way as I can on my phone.

jcass77:
I've been unable to reproduce the issue where pausing and tapping quickly in different places causes the 'flick' behaviour. I've tried on my desktop and mobile using local files and streaming media.

Looking at the code, there are only two locations where the timer can start moving again:

1. Mopidy triggers a 'seeked' event, and the playback state is 'play' (not 'pause').
2. the webclient does ``setPlayState(true)``, which should only happen in response to the user clicking 'play' again.

https://github.com/pimusicbox/mopidy-musicbox-webclient/blob/c47ddd316cfed0bc0f552eef00246f0da80a68e7/mopidy_musicbox_webclient/static/js/progress_timer.js#L95-L98 should also ensure that there is only ever one valid target position that the timer should be aiming for, so not sure why it would keep repeating the cycle as you described.

So I'm stumped. @kingosticks do you want to try debugging this on your end or do you think this scenario is unlikely enough to occur with normal usage that we can push forward?

https://github.com/pimusicbox/mopidy-musicbox-webclient/commit/c47ddd316cfed0bc0f552eef00246f0da80a68e7 now also hides the timers and slider handle when playing streams where the track length is Infinity. I think this looks a little more refined than just having a disabled slider stuck at '0'.

jcass77:
And: have you perhaps tried incrementing the version number in the manifest file and cleaning the cache on the mobile browser to ensure it is using the latest version of all of the scripts?

jcass77:
I've been playing around with this for the last few days and it seems to be working well.

The last two commits fixed an issue where the timer would sometimes not start, and it also adds support for streams (i.e. infinite duration).

Any objections if we merge at this point?

kingosticks:
Sorry I forgot to reply regarding the weird issue I found. It's probably not something people will try and do so not worth blocking this. 

It's possible it was something weird, I tried it with streams and it seemed to work as expected without the extra commit..  Let me try again to test it on my phone one last time, for sanity if nothing else.

jcass77:
...merged with latest 'develop' to resolve conflicts.

kingosticks:
Seems perfect.  Made sure to clear the cache this time and couldnt reproduce what I found before. I think you can probably have rebased this on develop to avoid the extra merge commit. Very happy to merge.

jcass77:
Let me rebase quick and I'll do the merge.

PR #165, opened 2016-02-14
Fix scrolling in full screen for Chrome and Safari.

### LINKED ISSUE #53: Problem with fullscreen mode

jcass77:
I've simplified the full-screen toggling scripts, which seems to work more consistently on the browsers that I tested and also fixes #53.


mopidy_musicbox_webclient/static/js/gui.js
```diff
@@ -308,43 +308,35 @@ $(document).bind("pageinit", function() {
...
-function enterFullscreen() {
+function toggleFullscreen() {
     if (isMobileSafari) { alert ("To get this app in Full Screen, you have to add it to your home-screen using the Share button."); exit(); }
-    var elem = document.querySelector("#page");
-    elem.onwebkitfullscreenchange = onFullScreenEnter;
-    elem.onmozfullscreenchange = onFullScreenEnter;
-    elem.onfullscreenchange = onFullScreenEnter;
-    if (elem.webkitRequestFullscreen) {
-        elem.webkitRequestFullscreen(Element.ALLOW_KEYBOARD_INPUT);
+    if (!document.fullscreenElement &&    // alternative standard method
```
kingosticks: Is this big if to handle different browsers? If so, am surprised there's no jquery support for that.

jcass77: I couldn't find any direct support for fullscreen mode in jQuery. There are a bunch of jQuery plugins that does the same thing, but I'm a little reluctant to rely on those as a rule - maintenance for many of them appear to be spotty at best.

kingosticks: I agree with that.

kingosticks:
I added another question but I can't see it now. Must not have posted. 

What was the actual problem? I can't see where the functional change is.

jcass77:
I think it was a combination of using the element ```document.documentElement``` instead of ```#page``` and dropping the ```Element.ALLOW_KEYBOARD_INPUT``` parameter (which doesn't work in Safari).

kingosticks:
Appears to work fine for me now in Android v5.

Issue #99, opened 2015-04-24
Track artist is not displayed in browse results


kingosticks:
Originally posted at https://discuss.mopidy.com/t/artist-name-not-displayed-in-several-screens/682

>I'm just in the process of upgrading from musicbox 0.53 to 0.6 and noticed that the artist name is no longer displayed in the "Queue" screen. 
While I quite like the additional line showing the album name and album artist, the individual artists of the tracks are missing. This makes identifying the songs rather difficult.

kingosticks:
On further inspection, it seems that in the playlist and search pages, if there is just a single track from an album (and therefore no grey album row) you are shown the artist name alongside the track name. If you play from these pages then this is displayed the same way in the queue. So this is  intentional behaviour. 

Browse works somewhat differently.  Here we are returned lightweight Ref objects which don't contain the artist/album metadata. So you only get the track name. But since Mopidy v1.0+ is moving towards using Refs everywhere we'll need to handle this anyway. I guess we need to do lookups on all the returned results...

jcass77:
I have a question on this, having recently worked on refactoring the code responsible for displaying list of tracks as part of the search results and the 'now playing' queue.

Although I appreciate that it would be nice to have all of the various lists look and behave in an identical way, is the intention of using Refs not specifically to improve performance? I did a quick test and doing lookups on large numbers of track Refs does cause a noticeable delay during which the browser becomes unresponsive, even when using promises.

Is an effective implementation of this not blocked by the ability to paginate results as described in https://github.com/mopidy/mopidy/issues/1392 and https://github.com/mopidy/mopidy/issues/733?

In the interim, using something like [mopidy-local-sqlite](https://github.com/mopidy/mopidy-local-sqlite) might be a more effective means of browsing large collections of albums and artists.

kingosticks:
Yes, `Refs` are a big improvement when you can save transferring lots of metadata around *that you don't care about*. I think the idea here was saying that maybe we do care about this rich metadata when browsing. You could argue that for the `local` backend, we do/should care but for the `file` backend we don't. 

I'm not sure I understand how the user's particular local backend affects us.

PR #169, opened 2016-02-16
Refactor cover art retrieval

### LINKED ISSUE #128: Album art is not displayed
There are failing requests to last.fm for getting the album art. The returned message is always like:
```
    jsonp144560929891147({"error":6,"message":"Album not found","links":[]})
```
This seems to be the case because the request is missing the required artist name:
```
    http://ws.audioscrobbler.com/2.0/?artist=&album=The+Better+Life&method=album.getInfo&api_key=...&callback=jsonp144560929891147&format=json
```
Probably this was caused by a last.fm API change of [album.getInfo](http://www.last.fm/api/show/album.getInfo).

jcass77:
I've refactored the Javascript routines for retrieving album covers:

- Optimised to pass track URI's instead of full album models.
- Now uses the new ``mopidy.library.getImages`` convention before falling back to the deprecated ``album.images`` approach.
- Now also falls back to using track artist when doing lookups on last.fm if album artist is not available (fixes #128).

Just for the avoidance of doubt, the order of precedence is: ``mopidy.library.getImages`` -> ``album.images`` -> last.fm

### DIFF
--- a/mopidy_musicbox_webclient/static/js/functionsvars.js
+++ b/mopidy_musicbox_webclient/static/js/functionsvars.js
@@ -343,10 +343,10 @@ function resultsToTables(results, target, uri) {
                     }
                     newalbum = [];
                     newtlids = [];
+                    if (results[i].album) {
+                        coversList.push([results[i].uri, i]);
+                    }
                 } //newalbum length
-                if (results[i].album) {
-                    coversList.push([results[i].uri, i]);
-                }
             } //albums name
         }
     }
--- a/mopidy_musicbox_webclient/static/js/functionsvars.js
+++ b/mopidy_musicbox_webclient/static/js/functionsvars.js
@@ -224,7 +224,7 @@ function resultsToTables(results, target, uri) {
 
     var newalbum = [];
     var newtlids = [];
-    //keep a list of albums for retreiving of covers
+    //keep a list of track URIs for retrieving of covers
     var coversList = [];
     var nextname = '';
     var count = 0;
@@ -343,12 +343,9 @@ function resultsToTables(results, target, uri) {
                     }
                     newalbum = [];
                     newtlids = [];
-                    if (results[i].album) {
-                        coversList.push([results[i].album, i]);
-                    }
                 } //newalbum length
                 if (results[i].album) {
-                    coversList.push([results[i].album, i]);
+                    coversList.push([results[i].uri, i]);
                 }
             } //albums name
         }
@@ -356,7 +353,7 @@ function resultsToTables(results, target, uri) {
     tableid = "#" + tableid;
     $(target).html(html);
     $(target).attr('data', uri);
-    //retreive albumcovers
+    //retrieve albumcovers
     for (i = 0; i < coversList.length; i++) {
         getCover(coversList[i][0], target + '-cover-' + coversList[i][1], 'small');
     }
--- a/README.rst
+++ b/README.rst
@@ -77,6 +77,7 @@ v2.2.0 (UNRELEASED)
 - Scrolling now works in full screen mode for Chrome and Safari as well. (Fixes: `#53 <https://github.com/pimusicbox/mopidy-musicbox-webclient/issues/53>`_).
 - No longer interferes with changes to Mopidy's volume levels that are triggered externally. (Fixes: `#162 <https://github.com/pimusicbox/mopidy-musicbox-webclient/issues/162>`_).
 - Volume slider now works with Mopidy-ALSAMixer again. (Fixes: `#168 <https://github.com/pimusicbox/mopidy-musicbox-webclient/issues/168>`_).
+- Now falls back to track artist if album artist is not available for rendering cover art. (Fixes: `#128 <https://github.com/pimusicbox/mopidy-musicbox-webclient/issues/128>`_).
 
 v2.1.1 (2016-02-04)
 -------------------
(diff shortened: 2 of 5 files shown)


jcass77:
I think I'm satisfied to merge this if there are no further comments or inputs received in the next 24 hours?

kingosticks:
Yeh I will give this a test later today.

mopidy_musicbox_webclient/static/js/functionsvars.js
```diff
@@ -343,12 +343,9 @@ function resultsToTables(results, target, uri) {
                     }
                     newalbum = [];
                     newtlids = [];
-                    if (results[i].album) {
-                        coversList.push([results[i].album, i]);
-                    }
                 } //newalbum length
                 if (results[i].album) {
-                    coversList.push([results[i].album, i]);
+                    coversList.push([results[i].uri, i]);
```
kingosticks: Did you mean to move this outside the `if`? In the case of there being only one track in the album, the cover art is not displayed. Or is it requested anyway to be used in another view? I admit I'm not 100% familiar with this code.

kingosticks: No, it appears it's not used anywhere else. All other places where we display cover art we seem to re-request the image anyway.

mopidy_musicbox_webclient/static/js/functionsvars.js
```diff
@@ -356,7 +353,7 @@ function resultsToTables(results, target, uri) {
     tableid = "#" + tableid;
     $(target).html(html);
     $(target).attr('data', uri);
-    //retreive albumcovers
+    //retrieve albumcovers
     for (i = 0; i < coversList.length; i++) {
         getCover(coversList[i][0], target + '-cover-' + coversList[i][1], 'small');
```
kingosticks: Out of scope for now maybe, but do you have plans to group these into a single `getImages` request at some point?

mopidy_musicbox_webclient/static/js/functionsvars.js
jcass77: I couldn't see why it was necessary to have the same statement inside the ``if``, and then again immediately outside it. Did I miss something there?

mopidy_musicbox_webclient/static/js/functionsvars.js
jcass77: The whole ``resultsToTables`` function is a little hard to follow and probably due for some refactoring (I'm not sure where to start though).

Until I have a better understanding of what it does and how it works: I'm not sure that a ``getImages`` will be useful anywhere else. But I did notice that there are a lot of calls go ``getCover`` for just about every navigation action in the client, so doing it once and caching the results for retrieval later on would most likely make more sense.

kingosticks: Just to be clear, I meant the optimisation of calling mopidy.getImages just once, but with all the covers to fetch. Currently we do individual calls for each cover.

mopidy_musicbox_webclient/static/js/functionsvars.js
kingosticks: Oh, I didn't realise it was previously in *both* places. Surely once, inside the if statement should be sufficient. I agree it's a beast of a function and I've always steered clear but looking last night it's not too bad. It's just trying to split the results up by album so it can insert an album row (but only in the case where there are multiple consecutive  tracks for the album).

mopidy_musicbox_webclient/static/js/functionsvars.js
jcass77: Oh now I get it. Event so, at the moment only the call from ``resultsToTables`` will have more than one URI available for calling `` mopidy.getImages`` on, so we can probably leave that for another PR as you suggested.

I'm still a little confused as to why the images are being retrieved here to start with, as some of the tables being rendered in the client don't even have the ability to display images as far as I can tell.

kingosticks: I didn't notice that, which ones don't display images?

jcass77:
Anything else that needs to happen here before merging?

mopidy_musicbox_webclient/static/js/functionsvars.js
kingosticks: I'd suggest moving this inside the if

mopidy_musicbox_webclient/static/js/functionsvars.js
jcass77: From what I can see, ```resultsToTables``` only shows images if an 'album header' forms part of the search results:

skitched-20160221-200155

...and I think that only happens when all of the tracks in the entire album are listed in either the search results or the play queue. Many of the extensions that I am using adds individual tracks to the queue however, so I don't often get to see album covers in the results tables.

skitched-20160221-200429

Probably nothing wrong with this, just a little surprising in terms of what I would expect personally.

mopidy_musicbox_webclient/static/js/functionsvars.js
jcass77: ...so the idea is that we only push an album cover if more than one track is being shown for a particular album?

Done in 7d680fe252b5d4e5c28619e8ba1e67705891a0c2!

mopidy_musicbox_webclient/static/js/functionsvars.js
jcass77: My bewilderment is probably related to https://github.com/pimusicbox/mopidy-musicbox-webclient/issues/99.

kingosticks: An album header should be shown when they are more than one tracks form the same album. It should not require all the tracks. I believe it has this behaviour to avoid looking stupid (and twice as long) when it is just a big list of random songs.

kingosticks:
If you are able to just double check that moving that if statement still grabbed covers for everything that wanted one .. then LGTM.

jcass77:
Yes I think we are good.

Issue #113, opened 2015-06-18
Javascript prompt doesn't show in some browsers


kingosticks:
By default, some Web browsers block pop-up prompt and confirmation dialog boxes as a security measure. This will break the save queue functionality (#106). We need to replace this stuff with the usual jquery popup as used elsewhere.

DavisNT:
I will look into this.

Do we have small jQuery popup anywhere? Should we make `prompt()` and `confirm()` popups the same size and style as "track context menu" popups?

kingosticks:
Not sure we gave any that take input. But yes, like the others

DavisNT:
I will **not** fix this (at least for now) as I don't see this as a critical issue (if somebody will want this feature he will disable popup blocker for Pi MusicBox URL) and there is big possibility that my fix won't get merged anyway.

Issue #173, opened 2016-02-18
Playlists refresh button no longer works


kingosticks:
It seems that that the playlists refresh button does not work. 

1. Select a playlist to display the tracks
2. Externally make a change to that playlist (e.g. add tracks to a Spotify playlist using play.spotify.com)
3. Click the refresh button and reselect the playlist
4. The new tracks are not visible in the playlist.

The only way I could see the new tracks was to restart Mopidy

This probably happened when I changed to use the `getItems` API. Need to make sure we invalidate the cache of tracks when the refresh button is clicked and also when we receive any sort of playlist changed type event from Mopidy.

PR #174, opened 2016-02-22
Various fixes to UI dialogs and buttons.

jcass77:
In this PR:

- replace the buttons on the 'Queue' pane with jQuery Mobile's ``control group``.
- replace Javascript dialog prompts with jQuery Mobile's ``popup`` - @kingosticks, is this what you had in mind for #113?
- fixes playlist refresh button.
- update save queue functionality: now saves to 'm3u' format and no longer uses deprecated ``PlaylistsController.filter`` function.

### DIFF
--- a/mopidy_musicbox_webclient/static/js/controls.js
+++ b/mopidy_musicbox_webclient/static/js/controls.js
@@ -207,39 +207,67 @@ function removeTrack() {
 }
 
 function clearQueue() {
-    mopidy.playback.stop();
-    resetSong();
-    mopidy.tracklist.clear();
-    resetSong();
+    mopidy.tracklist.clear().then(
+        resetSong()
+    );
     return false;
+}
+
+function savePressed(key) {
+    if (key == 13) {
+        saveQueue();
+        return false;
+    }
+    return true;
+}
+
+function showSavePopup(){
+    mopidy.tracklist.getTracks().then(function(tracks) {
+        if (tracks.length > 0) {
+            $('#saveinput').val('');
+            $('#popupSave').popup('open');
+        }
+    });
+
 }
 
 function saveQueue() {
     mopidy.tracklist.getTracks().then(function(tracks) {
-        if (tracks.length > 0) {
-            var plname = window.prompt("Playlist name:", "").trim();
-            if (plname != null && plname != "") {
-                mopidy.playlists.filter({"name": plname}).then(function(existing) {
-                    var exists = false;
-                    for (var i = 0; i < existing.length; i++) {
-                        exists = exists || existing[i].uri.indexOf("m3u:") == 0 || existing[i].uri.indexOf("local:") == 0;
-                    }
-                    if (!exists || window.confirm("Overwrite existing playlist \"" + plname + "\"?")) {
-                        mopidy.playlists.create({'name': plname, 'uri_scheme': "local"}).then(function(playlist) {
-                             playlist.tracks = tracks;
-                             mopidy.playlists.save({'playlist': playlist}).then();
-                             getPlaylists();
-                        });
-                    }
-                });
-            }
+        var playlistName = $('#saveinput').val().trim();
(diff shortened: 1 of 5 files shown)


mopidy_musicbox_webclient/static/js/controls.js
```diff
@@ -207,37 +207,68 @@ function removeTrack() {
 }
 
 function clearQueue() {
-    mopidy.playback.stop();
-    resetSong();
-    mopidy.tracklist.clear();
-    resetSong();
+    mopidy.playback.stop().then(
```
kingosticks: If we want we can safely omit the `stop`command since clearing does that for us.

mopidy_musicbox_webclient/static/js/controls.js
```diff
@@ -207,37 +207,68 @@ function removeTrack() {
...
+        }
+    });
+
+}
+
+function saveQueue() {
+    mopidy.tracklist.getTracks().then(function(tracks) {
+        var plname = $('#saveinput').val().trim();
+        if (plname != null && plname != "") {
+            mopidy.playlists.filter({"name": plname}).then(function(existing) {
```
kingosticks: I know you are just reusing the code that was there, but we should note that `filter` is deprecated and will be removed in Mopidy v3.0

mopidy_musicbox_webclient/static/js/controls.js
```diff
@@ -207,37 +207,68 @@ function removeTrack() {
...
     });
     return false;
 }
 
+function initSave(playlistName, tracks) {
+    $('#popupOverwrite').popup('close');
+    $('#popupSave').popup('close');
+    $('#saveinput').val('');
+    toast('Saving...');
+    mopidy.playlists.create({'name': playlistName, 'uri_scheme': "local"}).then(function(playlist) {
```
kingosticks: Again, not your doing, but using this the local uri scheme isn't right (anymore).

kingosticks:
This is much smarter and defiantly addresses #113. Really nice.

Sadly, the whole queue save thing is [quite broken](https://github.com/pimusicbox/mopidy-musicbox-webclient/issues/177) and I was unable to test out the overwrite popup.

mopidy_musicbox_webclient/static/js/controls.js
jcass77: Ok, let's try that.

mopidy_musicbox_webclient/static/js/controls.js
jcass77: I'll update this while I'm busy working in this area of the code.

mopidy_musicbox_webclient/static/js/controls.js
jcass77: On it.

jcass77:
I've pushed https://github.com/jcass77/Mopidy-MusicBox-Webclient/commit/27778751b72fbc9272e269dbb8ab22c68ad810e0 which addresses all of the comments above.

Since all of this is loosely related, I also took a stab at fixing #173 and #177 as well. The web client now listens for ``playlistChanged`` and ``playlistDeleted`` events as well, so it should rarely be necessary to use the manual refresh button. But leaving it in is probably a good idea in case there are some use cases that we have not thought of?

jcass77:
https://github.com/jcass77/Mopidy-MusicBox-Webclient/commit/24100a38618e9fb18e7dd58326629afc202c70d5 changes links to proper buttons, and replaces text with equivalent FontAwesome icons to make better use of real estate on small screens.

jcass77:
Rebased on develop to resolve merge conflicts.

jcass77:
I can't resist the temptation of merging this and closing three issues in one fell swoop.

@kingosticks, I **think** this resolves all of the playlist-related issues - please let me know if you come across anything else that requires attention in this area.

PR #178, opened 2016-02-29
Lint fixes

jcass77:
I ran the .css and .html files through tidy-html5, csslint, and uncss which identified a few duplicate IDs and small issues in the HTML, and a few unused styles.

Nitpicking I know, but I wasn't very confident about the state of ``webclient.css``. At least now we know it wasn't too bad to start with.


jcass77:
I also ran all of the Javascript files through jshint.

I'm going to go ahead and merge this, hoping that all of the changes are fairly innocuous apart from fixing a few warnings and script errors. If this breaks anything for anyone, please let me know.

mopidy_musicbox_webclient/static/index.html
```diff
@@ -350,11 +351,13 @@ <h3 id="h_artistname"></h3>
...
             <form>
                 <p>Search for artists, albums, or specific tracks.
-                <select id="selectSearchService"> <!-- data-native-menu="false"> <!-- multiple="multiple" data-native-menu="false">
+                <select id="selectSearchService"> 
+                    <!-- data-native-menu="false"> 
+                    multiple="multiple" data-native-menu="false">
                     <option data-placeholder="true">Choose services</option> -->
                 </select>
                 <input id="searchinput" placeholder="Search term" class="span2" data-clear-btn="true"
-                       onkeypress="return searchPressed(event.keyCode);" id="appendedInputButton" type="text"/>
```
kingosticks: I wonder why this had two IDs...

jcass77: ``appendedInputButton`` isn't referenced anywhere that I could find, so I'm hoping that this was just left over after a code refactoring exercise, since long forgotten.

mopidy_musicbox_webclient/static/js/gui.js
```diff
@@ -508,11 +509,13 @@ $(document).ready(function(event) {
 
     //navigation temporary, rewrite this!
     $('#songinfo').click(
-	function()
-    	 {return switchContent('nowPlaying')}   );
+	function() {
+        return switchContent('nowPlaying');
+    });
```
kingosticks: Better formatting but still not quite there.

kingosticks:
Good call on the merge though, I was going to suggest just to go ahead and do it. My only real comment is that adding the `<title>` tags is a bit misleading since they won't be used. But no big deal.

mopidy_musicbox_webclient/static/js/gui.js
jcass77: I'm trying out Sublime Text for the first time and didn't notice that it doesn't convert tabs to spaces by default.

I'll fix this and a few other spacing issues and just push it directly to develop.

jcass77:
According to [w3schools](http://www.w3schools.com/tags/tag_title.asp) *"Tip: If you omit the ``title`` tag, the document will not validate as HTML."*.

I just added it to satisfy tidy-html5, but not sure it makes an actual difference.

PR #180, opened 2016-03-02
Add tox environments for running linters

jcass77:
This PR contains the first part of integrating style and syntax checking for the HTML, CSS, and Javascript source files used in MMW as part of the build. Once we agree on the general approach, the next step will be to modify the travis build and then run the linters to fix violations for the rule sets that we agree on.

General principles:
- Avoid adding additional dependencies that developers would need to install in order to run the linters
- Keep it simple and try to avoid task runners and development workflow tools like grunt, gulp, etc.
- Use an approach that can be easily applied to the travis builds as well

The above however means making some concessions, more on that below...(please ignore the *.js changes - PyCharm insists on removing the superfluous whitespace characters and I could not get it to revert to develop).

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,0 +1,33 @@
+{
+  "name": "Mopidy-MusicBox-Webclient",
+  "version": "2.1.1",
+  "description": "Mopidy MusicBox web extension",
+  "main": "gui.js",
+  "directories": {
+    "test": "tests"
+  },
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1",
+    "eslint": "eslint mopidy_musicbox_webclient/static/js/**.js",
+    "csslint": "csslint mopidy_musicbox_webclient/static/css/**.css",
+    "tidy": "node tidy.js"
+  },
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/pimusicbox/mopidy-musicbox-webclient.git"
+  },
+  "author": "Wouter van Wijk",
+  "license": "Apache-2.0",
+  "bugs": {
+    "url": "https://github.com/pimusicbox/mopidy-musicbox-webclient/issues"
+  },
+  "devDependencies": {
+    "eslint": "latest",
+    "eslint-config-standard": "latest",
+    "eslint-plugin-standard": "latest",
+    "eslint-plugin-promise": "latest",
+    "csslint": "latest",
+    "tidy-html5": "latest"
+  },
+  "homepage": "https://github.com/pimusicbox/mopidy-musicbox-webclient#readme"
+}
--- a/tidy.js
+++ b/tidy.js
@@ -1,0 +1,88 @@
+var tidy = require('tidy-html5').tidy_html5
+
+var fs = require('fs')
+
+// Traverse directory structure looking for 'html' or 'htm' files.
+var getAllHtmlFilesFromFolder = function (dir) {
+    var filesystem = require('fs')
+    var results = []
+    filesystem.readdirSync(dir).forEach(function (file) {
+        file = dir + '/' + file
+        var stat = filesystem.statSync(file)
+
+        if (stat && stat.isDirectory()) {
+            results = results.concat(getAllHtmlFilesFromFolder(file))
+        } else {
+            var extension = file.substr(file.lastIndexOf('.') + 1).toUpperCase()
+            if (extension === 'HTM' || extension === 'HTML') {
+                results.push(file)
+            }
+        }
+    })
+    return results
+}
+
+// Read file contents.
+function readFiles (dirname, onFileContent) {
+    var filenames = getAllHtmlFilesFromFolder(dirname)
+    filenames.forEach(function (filename) {
+        fs.readFile(filename, 'utf-8', function (err, content) {
+            if (err) {
+                throw (err)
+            }
+            onFileContent(filename, content)
+        })
+    })
+}
+
+var util = require('util')
+
+// Trap stderr output so that we can detect parsing errors.
+function hook_stderr (callback) {
+    var old_write = process.stderr.write
+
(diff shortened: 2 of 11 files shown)


.editorconfig
```diff
@@ -0,0 +1,22 @@
+# This file is for unifying the coding style for different editors and IDEs
```
jcass77: The idea is that this could be used in editors like Sublime Text. But it is not essential and we can remove if needed.

package.json
```diff
@@ -0,0 +1,33 @@
+{
```
jcass77: Mainly we just need this to define the script targets so that we can do things like ``npm run eslint``.

tidy.js
```diff
@@ -0,0 +1,88 @@
+var tidy = require('tidy-html5').tidy_html5
```
jcass77: The tidy-html5 npm does not ship with a command line interface, so we need to wrap it in a script file :(

build_env.py
```diff
@@ -0,0 +1,12 @@
+from __future__ import absolute_import, division, print_function, unicode_literals
```
jcass77: We won't need this script if we could run the ``nodeenv -p --prebuilt`` command from tox. But this means installing node and npm every time that we run tox which is quite slow.

This script installs the node environment in the ``node_env`` folder the first time it is run, and then just re-uses it on subsequent calls.

tox.ini
```diff
@@ -13,11 +13,39 @@ commands =
...
+commands = flake8 {posargs:mopidy_musicbox_webclient}
+
+[testenv:eslint]
+whitelist_externals =
+    /bin/bash
+deps =
+    nodeenv
+skip_install = true
+commands =
+    python build_env.py
```
jcass77: ...this could be replaced with ``nodeenv -p --prebuilt`` but would be much slower.

tox.ini
```diff
@@ -13,11 +13,39 @@ commands =
...
+    bash -c '. node_env/bin/activate; npm run eslint'
```
jcass77: I could not get tox to accept ``. node/env/bin/activate`` or ``source node/env/bin/activate`` as valid commands. So need to pass this command string to bash unfortunately.

jcass77:
Forgot to add: our options are basically to try and [use npm as a build tool](http://blog.keithcirkel.co.uk/how-to-use-npm-as-a-build-tool/), or resort to something like grunt.

If using npm, the next question is how to install npm and node on the user's machine. [nodeenv](https://github.com/ekalinin/nodeenv) seems to be the most practical as it can just be installed with pip. It does not support Windows unfortunately but it appears that work on that has already started (https://github.com/ekalinin/nodeenv/issues/140)

[nvm](https://github.com/creationix/nvm) is also an option and available in the travis Python environment, but getting it installed will differ on each platform and we do not really need to test against different versions of node, so the version management aspect of that tool is a bit of an overkill.

And lastly, tox isn't really geared for all of this, so I had to add a few scripts and define a few ugly build targets to make it all work. But I **think** it is simple enough to consider, and it should be really easy to integrate with travis.

jcass77:
Just a note that I pushed https://github.com/jcass77/Mopidy-MusicBox-Webclient/commit/94aeb2f48528692f2b02f1edd34f6f9546c81e22 which pre-installs all of the required node.js packages in the node virtual environment.

jcass77:
Turns out that getting this into travis required only three lines of config: https://github.com/jcass77/Mopidy-MusicBox-Webclient/commit/82984d787f49c308d2f6399c68af53c258975b90

jcass77:
It's probably better to just rely on ``npm install`` for getting the dependencies in instead of having to maintain a separate ``requirements.txt`` file. https://github.com/jcass77/Mopidy-MusicBox-Webclient/commit/ece436532bd981a1936eef39df59ccefb112a1fe does that.

build_env.py
jodal: By default, tox caches it's environment in the `.tox` directory. If you pass `-r` to tox, it forces recreation of the env.

jcass77: Thanks for the pointers. It seems that running ``nodeenv`` with ``-p`` recreates the node environment every time. And trying to do ``nodeenv {dest_dir}`` causes and error if the environment already exists.

I did some digging and found that one can suppress errors in tox by adding a ``-`` prefix to a command. So at least now we don't need the ``build_env.py`` script anymore.

jcass77:
I think this is done. In summary we have two options for the tox command:

```
- nodeenv --prebuilt {toxworkdir}/node_env
bash -c '. {toxworkdir}/node_env/bin/activate; npm install; npm run <node_script>'
```
This creates just one node environment shared by all of the tox environments and it also persists between runs so is the fastest by far. Downside here is that we need to execute a bash command to activate the node environment and run the npm commands within that environment.
```
nodeenv -p --prebuilt
npm install
npm run <node_script>
```
This uses the path of the python virtualenv and the npm commands can be executed as-is. Downside here is that the node environment is created every time for every tox environment so this is quite a slow process.

I prefer the first option since the tox runs are much quicker. But the second will probably be more portable.

Any preferences or suggestions from anyone?

jcass77:
I think I am going to merge this later today and then run ``eslint --fix`` to clean up the JavaScript files.

Everyone ok with [JavaScript Standard Style](http://standardjs.com)?

jcass77:
...except we will be using four spaces for indentation instead of two.

.editorconfig
kingosticks: I don't think we need this.

jcass77: Agreed. I'll take it out before merging.

jcass77:
All of the style changes have been committed as part of 979dc028c788d776078d2eb67d82b480a429a7e2 - the travis build is green again.

PR #181, opened 2016-03-06
Don't create Mopidy models manually.

### LINKED ISSUE #172: Stop creating Mopidy models
In at least one place we locally create and populate instances of Mopidy's models ourselves. Previously this idea has led to bugs when we then tried to pass these models back to Mopidy, at which point they turned out to invalid/wrong format/junk. 

I think we now only pass `Ref`s to Mopidy so bugs like this are hopefully gone. However, I can't think of any good reasons why we'd need to create and populate these models ourselves locally and if it's no longer necessary (since we don't use them externally) we should aim to entirely remove this practice. 

One such example of this is at https://github.com/pimusicbox/mopidy-musicbox-webclient/blob/17c8920d6999ee0158054f5019f0156d58eab7cc/mopidy_musicbox_webclient/static/js/functionsvars.js#L241-L250

jcass77:
Fixes #172.

I also refactored the ``resultsToTables`` function to reduce the amount of code duplication and to avoid having to iterate through each list of album tracks twice to show the album artwork.


jcass77:
I added a few more helper functions to simplify the implementation of ``resultsToTables`` even further.

Should be ready to merge if somebody can do a quick check...

mopidy_musicbox_webclient/static/js/functionsvars.js
```diff
@@ -176,185 +176,156 @@ function albumTracksToTable (pl, target, uri) {
...
-        name = song.name
+        iconClass = 'trackname'
     }
-    songLi = '<li class="song albumli" id="' + liID + '">' +
-             '<a href="#" class="moreBtn" onclick="return popupTracks(event, \'' + uri + '\',\'' + song.uri + '\');">' +
+    // Play by tlid if available.
+    if (tlid) {
+        tlidString = '" tlid="' + tlid
+        tlidParameter = '\',\'' + tlid
+        onClick = 'return playTrackQueueByTlid(\'' + song.uri + '\',\'' + tlid + '\');'
```
kingosticks: Am I being dense or can't you use double quotes for the outer quotes here and save the less readable escaping?

kingosticks: Maybe that mixture of styles would be more confusing. Up to you.

mopidy_musicbox_webclient/static/js/functionsvars.js
```diff
@@ -176,185 +176,156 @@ function albumTracksToTable (pl, target, uri) {
...
+            }
+        }
+    }
+    return html
+}
+
+function isNewAlbumSection (track, previousTrack) {
+    // 'true' if album name is either not defined or has changed from the previous track.
+    return !track.album || !track.album.name || !previousTrack || !previousTrack.album || !previousTrack.album.name ||
+           (previousTrack && previousTrack.album && previousTrack.album.name && track.album.name !== previousTrack.album.name)
```
kingosticks: If we have got this far we already know `previousTrack && previousTrack.album && previousTrack.album.name`.

mopidy_musicbox_webclient/static/js/functionsvars.js
```diff
@@ -176,185 +176,156 @@ function albumTracksToTable (pl, target, uri) {
...
+    // 'true' if there are more tracks of the same album after this one.
+    return nextTrack.album && nextTrack.album.name && track.album && track.album.name && track.album.name === nextTrack.album.name
+}
+
+function validateTrackName (track) {
+    // Create name if there is none
+    var name = ''
+    if (!track.name || track.name === '') {
+        name = track.uri.split('/')
+        name = decodeURI(name[name.length - 1]) || 'Track ' + String(i)
```
kingosticks: What is `i` here?

mopidy_musicbox_webclient/static/js/functionsvars.js
```diff
@@ -176,185 +176,156 @@ function albumTracksToTable (pl, target, uri) {
...
+        previousTrack = track
+        track = results[i]
+        tlid = ''
+        if (i < length - 1) {
+            nextTrack = results[i + 1]
         }
-        // create album uri if there is none
-        if (!results[i].album.uri) {
-            results[i].album.uri = 'x'
+        if (track.__model__ === 'TlTrack') {
```
kingosticks: Is looking at the `__model__` property better than looking for the existence of a `tlid` property? It seems more brittle.

mopidy_musicbox_webclient/static/js/functionsvars.js
```diff
@@ -176,185 +176,156 @@ function albumTracksToTable (pl, target, uri) {
...
+            // Get track information from TlTrack instance
+            track = track.track
+            tlid = results[i].tlid
+            nextTrack = nextTrack.track
```
kingosticks: `nextTrack` is undefined when the queue has just one entry.

mopidy_musicbox_webclient/static/js/functionsvars.js
```diff
@@ -176,185 +176,156 @@ function albumTracksToTable (pl, target, uri) {
...
+            // Starting to render a new album in the list.
+            tableid = 'art' + i
+            // Render differently if part of an album
+            if (i < length - 1 && isMultiTrackAlbum(track, nextTrack)) {
+                // Large divider with album cover
+                renderAlbumInfo = false
+                html += '<li class="albumdivider">'
+                html += '<a href="#" onclick="return showAlbum(\'' + track.album.uri + '\');">' +
+                '<img id="' + targetmin + '-cover-' + i + '" class="artistcover" width="30" height="30" />' +
+                '<h1><i class="' + getMediaClass(track.uri) + '"></i> ' + track.album.name + '</h1><p>'
```
kingosticks: Do we know `track.album` and `track.album.name` are defined for use here?

mopidy_musicbox_webclient/static/js/functionsvars.js
```diff
@@ -176,185 +176,156 @@ function albumTracksToTable (pl, target, uri) {
...
-             '<a href="#" onclick="return playTrackByUri(\'' + song.uri + '\',\'' + uri + '\');">' +
-             '<h1 class="trackname">' + name + '</h1></a>' +
-             '</li>'
+             '<a href="#" onclick="' + onClick + '">' +
+             '<h1><i class="' + iconClass + '"></i>' + song.name + '</h1>'
+    if (renderAlbumInfo) {
+        songLi += '</p>'
+        songLi += renderSongLiTrackArtists(song)
+        if (song.album && song.album.name) {
+            songLi += ' / '
```
kingosticks: Shall we switch this for a `-`? I'm not sure why it's the same separator character, it's unclear.

kingosticks:
Looks much improved but a few niggles to address I think.

mopidy_musicbox_webclient/static/js/functionsvars.js
jcass77: One of the linters that I tried out did not like mixed used of single and double quotes. However eslint seems to handle this a little better so I'll make the changes everywhere.

mopidy_musicbox_webclient/static/js/functionsvars.js
jcass77: Ok, ``-`` it is.

mopidy_musicbox_webclient/static/js/functionsvars.js
jcass77: Right, thanks.

mopidy_musicbox_webclient/static/js/functionsvars.js
jcass77: ``i`` came out of a ``for`` loop - I missed this when I moved this to the helper function.

I'll just pass the track number in to the helper function so that we can retain the feature to rename unnamed tracks to 'Track n'.

mopidy_musicbox_webclient/static/js/functionsvars.js
jcass77: I actually wanted to perform the check on the object type. But that seems to resolve to ``Object`` at this point so did not turn out to be very helpful.

I'll just check for the existence of ``tlid`` for now.

mopidy_musicbox_webclient/static/js/functionsvars.js
jcass77: ...in the end, using a mixture of styles just becomes too hard to keep straight between the editor, the linter, and how the browser interprets the string. Think I am going to stick to just escaping the quotes for now.

mopidy_musicbox_webclient/static/js/functionsvars.js
jcass77: Fixed to always read ``nextTrack`` from ``results[i + 1]``.

Having just one track in the queue works for me with this fix?

jcass77:
All comments addressed as part of https://github.com/jcass77/Mopidy-MusicBox-Webclient/commit/b298854b3ca5766ddb8caf30a049b68a194db269.

kingosticks:
I am still not seeing the queue when it contains just one track. Do you see this behaviour?

kingosticks:
Right, forget that, single queue works just fine. I can only assume Chrome is mega aggressive at caching and I need to open the dev console (with cache disabled ticked) to reload properly.  Isn't that manifest thing meant to stop this happened? What prevents normal users having to do this?

kingosticks:
And that is pretty rhetorical to be honest. This all looks great to me.

jcass77:
I think the check to see if the results table contains ``TlTrack`` was bad. Please retry with https://github.com/jcass77/Mopidy-MusicBox-Webclient/commit/858686d98c5e0c5bc99ef4c8b8e6a0d84d1f5a17.

jcass77:
I also found Chrome to be very aggressive on the caching. It's probably a good thing as I think the other browsers just ignore ``mb.appcache`` in favour of their normal caching routines (i.e. Chrome acts more like a HTML5 application would than the others).

As far as I can recall I could always get Chrome to reload everything by changing the date or version number in the ``mb.appcache`` header.

For normal users: we just need to be sure to increment the appcache version numbers before each release.

kingosticks:
In an ideal world we'd test with some tracks without album fields but I'm happy with this as is. If you are too then feel free to hit the BGB.

jcass77:
I had a few tracks without album fields that I tested with - worked fine so that use case is covered.

I'll rebase and merge, thank you for reviewing.

kingosticks:
Oh great stuff, nice one.

Issue #96, opened 2015-04-12
Add queue management to all hierarchy levels of local files


jo-me:
It would be good to have the queue management popup (add to queue, play next, add to end of queue,...) to all levels of the hierarchy.

That way, you could for example add all songs of a certain album, artist, etc.

This would be especially nice in the folder hierarchy as it would then allow to add all files from a folder somewhere in the hierarchy. This makes sense for albums that have several subfolder (CD1 & CD2) and it would even allow adding ALL files to the tracklist by executing it on root folder level:

![image](https://cloud.githubusercontent.com/assets/10351272/7107016/cb143e9c-e155-11e4-8347-de56c2ab95a9.png)

jcass77:
I'm not quite sure how one would go about implementing this cleanly and in a way that will be portable across all backends.

At the moment, the options are to use one of the following methods to build the tracklist:
- ``LibraryController.browse`` - provides a listing of items contained in the URI root, but not recursively as far as I know.
- ``LibraryController.search`` - it may be possible to call this with e.g. ``{uris=['file:///MusicBox']}``, but not all backends support searching.

These are just some initial thoughts...

ricstott:
I am guessing there hasn't been much need for this, given it is an oldish post? It is something I would love to see (and would totally help with if I had the know-how). I would have thought a lot of people organize music locally using a several directory layers? I certainly do.

kingosticks:
Are you using the latest version? Unless I misunderstand I think @jcass77 already implemented this.

Issue #69, opened 2015-02-26
Add whole playlist, folder or album to the queue.


EHLOVader:
The triple dot menus don't appear on the above items anywhere. It would be nice to add entire lists of things at once to the queue.

kingosticks:
Doesn't clicking the 'above item' do that?

EHLOVader:
I see now, you can use "Add all to bottom of queue" I think it would be nice to have the menus attached to the side of the parent item. OR in the header of the parent item.

Also perhaps wording might help. Is All the entire page, or just this album in the artist or search results?

![Add all to bottom of queue](https://i.imgur.com/sfxnhRcm.png)

I would like to see the dots over here instead. 

![Playlist item](https://i.imgur.com/OCKRLfNm.png)

and possibly something up here in the album header.

![Album header](https://i.imgur.com/QRLalfDm.png)

jcass77:
The convention that we are following at the moment is that all tracklist actions should originate from the popup menus linked to a track, rather than having additional menus in the UI that users have to go and look for on every pane.

Once https://github.com/pimusicbox/mopidy-musicbox-webclient/pull/193 is merged, we'll have all of the actions available wherever a track is rendered as part of a list, so this functionality should be a lot more consistent.

The only feature that I think you may be missing in the latest release is the ability to add just the tracks of a specific album to the queue. This can be done with two clicks if you select the 'Show Album...' option, and then 'Play All' on the screen that follows. I'd prefer using this approach rather than adding yet another option to the popup menus, which are becoming rather long and cumbersome as it is.

I'm closing this as I think all of the requests have been addressed (if not also provided for in the code since the issue was logged originally).

PR #182, opened 2016-03-10
Refactor code for displaying track lists.

### LINKED ISSUE #99: Track artist is not displayed in browse results

### LINKED ISSUE #126: Can't use contextmenu in browse pane
Nothing happens, when I click on the 3dot-button. Is this a common issue and does somebody know what the problem is? I'm on version 0.6 and didn't upgrade the webclient or something like that.

jcass77:
This PR represents the start of an effort to consolidate how lists of tracks are retrieved and displayed in the UI:

- fixes an issue that caused the popup menu in the browse pane to render differently than the others, or not at all. Fixes #126.
- further refactoring of ``renderSongLi``. There's more work to be done here as the various ``play...`` functions can also probably be refactored and consolidated. I'd eventually like to get to a point where we only have one list of tracks that can be cached and used in all of the panes. But that is a PR for another day.
- fixed a small issue with the default images for artists and albums sometimes not displaying if they cannot be retrieved from last.fm.
- browsing using the ``local`` or ``file`` extensions will now also render album and artist details, which fixes #99. It works well but the performance impact is noticeable for lists containing more than a hundred tracks or so. We might eventually either make this a configuration setting, or include a button on-screen that the user can use to toggle the details for any ``TrackRefs``. Ultimately, this could also be used to sort by artist or album on the fly. 

The rules for displaying album details are currently as follows:
1. if we only have one track in the list, then show its album details
2. if we have more than one consecutive track from the same album, then add an album header above the tracks. Album details are not repeated for each of the album's tracks.
3. if we do not have any album info available or consecutive tracks are from different albums, then display a think grey divider between the tracks.

This PR also paves the way for fixing #69 and #96.


jcass77:
Just a note that I have pushed some more change which ensures that the loading popup is displayed until after all of the album details have been retrieved.

Also fixes an issue where the album details for the first track in the list was not being rendered.

jcass77:
I've pushed https://github.com/pimusicbox/mopidy-musicbox-webclient/commit/a00754dd389139fec1812687f7092e852e359b4c which ensures that each track name is proceeded with an icon (except if part of a multi-track album) and a space character.

jcass77:
https://github.com/jcass77/Mopidy-MusicBox-Webclient/commit/60041c142ed62e5134cdc0bf19b219488b463abf refactors and consolidates how the 'Back' button is rendered at the top of track lists, and how messages are displayed when a folder does not contain any tracks.

https://github.com/jcass77/Mopidy-MusicBox-Webclient/commit/1c1fbee137a3cb5c3a466c215e9589bcf04cbe0b makes it more obvious if searches came up empty by showing 'No tracks found...' in the results table.

I'll run a few more tests and merge this tomorrow (Sunday) morning if there are no further inputs from anyone.

kingosticks:
I'm away this weekend so I cannot try this.

jcass77:
https://github.com/pimusicbox/mopidy-musicbox-webclient/commit/5437205e4df51eece55d5034bbe11a73799b5d10 optimises the jQuery selectors to use identifiers instead of attributes. Lists with roughly 2 000 tracks are now rendered about 80 times faster.

Also fixed a small issue with play icons not being updated in the ``#albumstable``, and replaced string references with their JavaScript constants.

jcass77:
I've been using this for a few days without any apparent issues. I'll merge later tonight if there are no objections.

PR #183, opened 2016-03-12
Add Javascript test and test coverage frameworks.

jcass77:
Adds Karma + Mocha + Chai testing frameworks, and integrates everything using tox.

This PR is mainly about agreeing on how continuous integration should be done, and it only includes half a test case to show that everything works.

Points for discussion:
- I haven't been able to find a way to integrate multiple coverage reports. So only the javascript reports will be pushed to coveralls.io. I assume it should be possible to merge ``lcov.info`` (produced by karma-coverage) and ``.coverage`` (produced by coverage.py) and push everything to coveralls.io as part of a single report?
- when writing the actual tests, I'm assuming that we'll just 'browserify' the test cases instead of introducing polyfils and shims in the actual JavaScrip source.
- I assume that running the tests using `PhantomJS` is not as good as an actual browser. And we can use `Firefox` on travis. But the way it is now means that we are not making any assumptions as which browsers developers will have available on their systems.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,7 @@
 MANIFEST
 build/
 cover/
+coverage/
 coverage.xml
 dist/
 docs/_build/
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,7 @@ python:
 env:
   - TOX_ENV=py27
   - TOX_ENV=flake8
+  - TOX_ENV=test
   - TOX_ENV=eslint
   - TOX_ENV=csslint
   - TOX_ENV=tidy
@@ -19,4 +20,6 @@ script:
   - "tox -e $TOX_ENV"
 
 after_success:
-  - "if [ $TOX_ENV == 'py27' ]; then pip install coveralls; coveralls; fi"
+  # TODO: find a way to combine .py and .js coverage reports.
+  # - "if [ $TOX_ENV == 'py27' ]; then pip install coveralls; coveralls; fi"
+  - "if [ $TOX_ENV == 'test' ]; then cat ./coverage/lcov.info | ./node_modules/coveralls/bin/coveralls.js; fi"
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -1,0 +1,81 @@
+// Karma configuration
+
+module.exports = function (config) {
+    config.set({
+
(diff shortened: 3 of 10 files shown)


jcass77:
I've updated this PR with a few more commits as I'm slowly getting to grips with UI development and JavaScript unit testing.

1. moves functions in images.js to an object literal. As far as I can tell there is no getting around modularising the codebase if we want to develop and run unit tests. This is mainly because mocking libraries like [sinon](http://sinonjs.org) expect stubs and mocks to be created with ``mock(obj)`` or ``stub(object, "method")``.

2. as soon as we start to modularise, we're faced with having tests that only run in node.js, and production code that only works in browsers. As MMW serves as a browser-based UI, I think it makes sense to use [karma](https://karma-runner.github.io/0.13/index.html) to run all tests in a browser environment instead. Although we currently rely on PhantomJS for running tests, we could potentially look at leveraging [browserstack](https://www.browserstack.com) to troubleshoot and test corner cases on multiple platforms.

3. (2) above means that we need to use [browserify](http://browserify.org) to package everything up for karma and browser environments.

4. (3) implies that we ultimately need a build step that does ``./node_modules/.bin/browserify mopidy_musicbox_webclient/static/js/*.js -o mopidy_musicbox_webclient/static/js/bundle.js``, and then just include ``bundle.js`` in the html files and mb.appcache. The upside is that we can then minimise everything before shipping. The downside is requiring developers to have node installed on their development machines, and adding build script to ``package.json``, but at least we can then get rid of ``nodeenv`` which should make tox runs much faster.

5. as modularising the code will be a gradual process, we'll need workarounds like including the following at the end of our script files until we can run everything through browserify:
```Javascript
if (typeof exports !== 'undefined') {
    if (typeof module !== 'undefined' && module.exports) {
        module.exports = coverArt
    }
}
```
...and also loading all of the production code (``'mopidy_musicbox_webclient/static/js/**/*.js'``) into the karma run until we can ``require('./just_the_libraries_we_need.js')`` for test cases.

But that's it. If everyone can live with the above then I think we have a way of moving forward, increasing test coverage, and modularising the MMW codebase to make it easier to maintain in future.

P.S. The first test coverage report is available at [https://coveralls.io/builds/5432742](https://coveralls.io/builds/5432742).

PR #186, opened 2016-03-21
Convert to model view controller (MVC) architecture.

jcass77:
My apologies in advance for the large PR. Before taking this further I'd like the Mopidy core developers to do a sanity check on this approach please.

This is an effort to reduce the reliance on JavaScript, by introducing a Python model that can be accessed from JavaScript using web sockets (exactly like mopidy.js).

- introduces 'webclient.js', which is based on ``mopidy.js`` and can be used to send and receive web socket messages to / from ``mopidy_musicbox_webclient.Webclient``. 
- at the moment, the ``mopidy_musicbox_webclient.Webclient`` model just contains a few proof of concept methods for adapting MMW functionality based on what has been configured in ``mopidy.conf``.
- the PR also updates the Tornado request handlers to handle a few redirects, and allow MMW to accept web socket requests.
- combines test coverage reports for Python and JavaScript code before pushing to coveralls.io
- new npm scripts for browserifying and minifying MMW scripts. We're only processing ``webclient.js`` at the moment but the idea is that all of the scripts will eventually be contained in a ``bundle.js`` once everything has been modularised.

Questions:

1. can ``mopidy_musicbox_webclient.Webclient`` in ``webclient.py`` be considered thread-safe if it references the Mopidy core? I thought this would be fine seeing as everything is wrapped in a ``mopidy.http.HttpFrontend``, or would we need to rely on threaded RPC calls here as well?

2. Does this approach make sense, or is there a more lightweight approach to implementing MVC that anyone knows of?

Next steps: if there are no major issues or objections then I will work on writing more test cases for all of these changes before merging.


kingosticks:
> This is an effort to reduce the reliance on JavaScript

This is a *lot* of extra complexity and an entire extra new intermediate interface to maintain. Is this worth it? Can you give some concrete examples of the benefits this brings?

jcass77:
Yes, the complexity certainly *felt* high getting this all in. But looking back we're pretty much just piggy-backing on ``mopidy.js`` and ``mopidy.http.WebSocketHandler`` - so hopefully we won't have to maintain much beyond just keeping an eye on those.

The benefit really is the ability to move most of the application logic that is not strictly UI-related into Python, where it is easier to maintain, organise, and test in my opinion.

- Dealing with large tracklists is problematic on the client side, which feels like something that could probably benefit from being able to use generators.
- Although one could easily go overboard, having access to everything on PyPI could make it easier to implement things like caching and retrieving cover art and album info from other sources, etc. so we'll gain a lot of flexibility if we have direct access to all of the available packages via a Python model.
- There are some hard-coded configuration features that should ideally be moved out to ``mopidy.conf`` (e.g. https://github.com/pimusicbox/mopidy-musicbox-webclient/blob/develop/mopidy_musicbox_webclient/static/js/functionsvars.js#L77-L110). Doing this in Javascript is difficult because we don't have access to the mopidy configuration (as far a I know?). I also had it so that the entire configuration is stored in a cookie, but that seems wasteful.
- It might be useful to have access to Python's introspection features to figure out if a backend for example has searching implemented. The alternative is to have Mopidy implement a bunch of ``has_`` and ``can_`` methods so that we can determine which backends support what so that we can have the UI respond accordingly.

In general MVC feels easier to maintain over the long term, compared to having to keep track of the various JavaScript workarounds and global variables that Tornado injects in the html templates. But I could be wrong, and will defer to anyone that has experience with these things.

jcass77:
Just saw that there are some long term plans to simplify Mopidy.js (third bullet in https://github.com/mopidy/mopidy/issues/1460) by removing 'bane' and 'when'.

Since Webclient.js (yes we need a better name) emulates Mopidy.js, we should be able to benefit from any improvements made to that project.

kingosticks:
> Dealing with large tracklists is problematic on the client side, which feels like something that could probably benefit from being able to use generators.

I don't follow. Are you talking about python generators here? For pagination? Problematic, as in slow?

> Although one could easily go overboard, having access to everything on PyPI could make it easier to implement things like caching and retrieving cover art and album info from other sources.

Between Mopidy providing coverart and what we get from last.fm - are we not sorted in this department? This seems hypothetical.

> There are some hard-coded configuration features that should ideally be moved out to mopidy.conf 

Is any of that useful for users to configure? The less configuration options available the better.
 
> It might be useful to have access to Python's introspection features to figure out if a backend for example has searching implemented. 

I think this is something that should be provided to us by the Mopidy API. But I don't think there is anything planned in this department. We should look at proposing it.

My big concern is that any extra interface creates an additional point of failure. Keeping an eye on things is much easier said than done. And while the Mopidy API is great at remaining compatible between versions, I'm not sure how solid something like this would be. And then we have our own backwards compatibility problems to think about. There is something to be said for just consuming what we are given.

Saying all that, I do see merit in being able to do more server-side things and the websocket events are nice to work with. Events aside though, you can probably make a more conventional templating approach quite neat. I'm just not sold on this.

jcass77:
Thanks for the feedback @kingosticks. I've been thinking about this some more and can perhaps provide additional context below.

>> Dealing with large tracklists is problematic on the client side, which feels like something that could probably benefit from being able to use generators.

> I don't follow. Are you talking about python generators here? For pagination? Problematic, as in slow?

Yes to all of the above. One could argue that having thousands of tracks in a single list is not a valid use case, but it would be nice if MMW could handle this more elegantly. Chrome seems to do better than the other browsers at dealing with this. Firefox eventually displays a popup warning message that the script is taking longer than expected, which can be a little jarring for the user.

>> Although one could easily go overboard, having access to everything on PyPI could make it easier to implement things like caching and retrieving cover art and album info from other sources.

> Between Mopidy providing coverart and what we get from last.fm - are we not sorted in this department? This seems hypothetical.

Yes I will concede this is hypothetical. Caching is the one area that we might want to look into, as well as reducing the number of track lists that we keep in memory (MMW pretty much copies the same list to every pane at this stage). I think this will be easier to rectify server-side.

>> There are some hard-coded configuration features that should ideally be moved out to mopidy.conf

> Is any of that useful for users to configure? The less configuration options available the better.

Because MMW is a 'general purpose' frontend, users have the freedom to plug in any backend that they want and expect MMW to be able to serve it up. In the past MMW made assumptions about which back ends are out there, what they should be called, and if they can be searched. This might not work for everybody all of the time.

The way MMW works at the moment, users should be able to configure human-friendly names to display on-screen at the very least.

>> It might be useful to have access to Python's introspection features to figure out if a backend for example has searching implemented.

> I think this is something that should be provided to us by the Mopidy API. But I don't think there is anything planned in this department. We should look at proposing it.

Agreed. To be realistic however, the Mopidy issue list is almost 200 items deep and many of those will have a higher priority than this. I know PRs are welcomed but frankly struggle to keep up with and contribute to the core in addition to working on multiple extensions. Perhaps we just need an interim solution until we can come up with a generic approach for doing extension feature detection in the core?

> My big concern is that any extra interface creates an additional point of failure. Keeping an eye on things is much easier said than done. And while the Mopidy API is great at remaining compatible between versions, I'm not sure how solid something like this would be. And then we have our own backwards compatibility problems to think about. There is something to be said for just consuming what we are given.

Yes I think it comes down to just deciding if we should leave good enough alone. I personally think we are fairly safe in relying on the Mopidy.js approach. If it ever broke, then *none* of the HTTP clients would work anymore. But I could be wrong.

> Saying all that, I do see merit in being able to do more server-side things and the websocket events are nice to work with. Events aside though, you can probably make a more conventional templating approach quite neat. I'm just not sold on this.

I like the idea of just using templates but am not sure how to apply this in MMW:

- we're already using web sockets and promises for Mopidy models, and will now introduce another template-based approach for MMW models.
- part of the reason MMW feels so responsive is because it loads most of what it needs into index.html when it starts up, and then just dynamically shows or hides sections using jQuery panes. Tornado organizes apps using request handlers and templates. If we just have one big handler for index.html then we are back to not being able to modularize our code properly, and end up with one massive dictionary for all content. If we organize it properly into different URL's, handlers, and templates, then every user action will trigger a request / response cycle which will probably slow things down a bit.

So I am unsure where to go from here. Perhaps we just leave it as is and not embark on moving stuff to the server side. There are some other useful elements in this PR that I can strip out and resubmit.

jcass77:
I started converting this to a Tornado template-based approach and need some help with another issue:

We use ``mb.appcache`` to cache everything in order to improve performance and startup times on especially mobile devices. The way I understand it, this means that Tornado will only render the html and JavaScript templates the first time they are retrieved, so if a template value changes (e.g user configures AlarmClock) the boolean variables will not be updated, causing unexpected behavior in MMW.

Our options are probably to:

1. Disable mb.appcache for the affected files (will eventually be everything but the vendor libraries and a few icons)
2. Resort to weird hacks like http://stackoverflow.com/questions/9287044/my-html5-application-cache-manifest-is-caching-everything/9287604#9287604

Then, just a closing comment that debugging js files that have gone through a Tornado handler takes some getting used to. You eventually have lots of magic strings showing up in the code that you keep needing to refer back to the template to to see where they came from.

jcass77:
Closing this proof of concept. Let's review if / when moving substantial portions of the application logic server-side ever becomes necessary.

PR #187, opened 2016-03-27
Convert to MVC-based architecture (Tornado template alternative)

jcass77:
So...this is the Tornado template-based alternative to https://github.com/pimusicbox/mopidy-musicbox-webclient/pull/186.

I'll readily admit that I have absolutely no experience using Tornado or templates, so please review the code to see if I am on the right track.

- At the moment we are using two different approaches for getting data from the model into the template: (a) storing them in ``<body data-xxx>`` tags and then retrieving with ``$(document.body).data('xxx')``, and (b) just injecting JSON strings directly into the JavaScript using Tornado's ``{{ xxx }}`` token format. I think (a) is neater but this will soon become impractical as we add more content, so we'll probably have to standardise on (b).
- I'm assuming that we'll need a different request handler for each JavaScript file so that we *just* make the Mopidy calls that are required to serve that script. Having a 1:1 mapping between Python models, request handlers, and JavaScript files seems crazy to me, so I am sure that I am doing this wrong - please advise? Worst case we end up having a different .js file for each method call?

Please ignore the test failures for now: due to the fact that we are injecting Tornado template tokens, parsing the raw JavaScript files causes eslint parsing errors (fun). I'll see if I can find a workaround for this in due course.

I still like the websocket-based approach better =)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,7 +8,7 @@
 MANIFEST
 build/
 cover/
-coverage/
+.karma_coverage/
 coverage.xml
 dist/
 docs/_build/
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,17 @@
-sudo: false
+sudo: required
+dist: trusty
 
 language: python
 
 python:
   - "2.7_with_system_site_packages"
+
+addons:
+  apt:
+    sources:
+      - mopidy-stable
+    packages:
+      - mopidy
 
 env:
   - TOX_ENV=py27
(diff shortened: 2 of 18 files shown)


jcass77:
I couldn't find a sensible way to incorporate Tornado templates in MMW. 

As this PR contains some other useful refactoring and additional tests, I've merged everything except the template changes to develop in 411a3781fdab270984299e47f136abec8395ff57.

PR #189, opened 2016-04-02
Upgrade media progress timer. Modularise progress_timer.js.

jcass77:
- Upgrades media progress timer to version 3.0.0.
- Modularises progress_timer.js (now called synced_timer.js) so that it works in node.js and browsers.
- Refactors and simplifies the MMW progress timer to just act as a wrapper to @adamcik's `timer.js`.
- The workaround for https://github.com/adamcik/media-progress-timer/issues/3 is now contained in `synced_timer.js` so that we can use and un-patched version of the vendor library.
- Unit tests for `SyncedProgressTimer`.
- Small fixes and tweaks to some of the other unit tests.


jcass77:
I've rebased and made a few more changes to switch library.js and images.js to the UMD loader pattern as well, and changed a few function signatures to start moving away from using global variables everywhere.

My apologies for the large commit. The test coverage for these changes is fairly high so hopefully there shouldn't be any surprises.

I'm going to go ahead and merge this, but please let me know if anybody has any comments on this PR and I will gladly circle back to incorporate.

jcass77:
I've fixed a few more issues and added a load of tests in f43a9a7afaffc50e3abe278a3df36203f8ca7308, just in case anyone is following along.

PR #190, opened 2016-04-07
Retrieve images using MusicBrianz ID

jcass77:
This represents an effort to make sure that requests to last.fm for album covers and artist images will always be unique:

- Now uses unique MusicBrainz ID whenever it is available. This should be more reliable than simply relying on album and artist names.

- `getCoverFromLastFm` now also requires that *both* the album and artist names be supplied if the MusicBrainz ID is not available.

- In the absence of a MusicBrainzID, `getArtistImage` could potentially still retrieve the wrong artist image from last.fm (e.g. a different artist with the same or a similar name). I'm not sure if there is away around this as the [last.fm API](http://www.last.fm/api/show/artist.getInfo) does not provide for anything else at this stage.


jcass77:
137593c586f2291c2ff88ca593d7e489e288fc80 fixes last.fm call signature.

kingosticks:
Oh nice. Now if I could just remember my test case where the wrong cover was requested.

mopidy_musicbox_webclient/static/js/images.js
```diff
@@ -61,47 +64,92 @@
...
+                console.warn('Not enough tag information available to retrieve album cover from last.fm, using default')
+            }
+        },
+
+        getArtistInfo: function (track) {
+            var artistName = ''
+            var musicBrainzID = ''
+
+            if (track && track.album && track.album.artists && track.album.artists.length > 0) {
                 // First look for the artist in the album
```
kingosticks: Out of scope for this PR. But does this make sense do you think? Seems that for various artists compilations this does the wrong thing and it should use `track.artists` first.

mopidy_musicbox_webclient/static/js/images.js
```diff
@@ -61,47 +64,92 @@
...
                 return
             }
-            this.lastfm.artist.getInfo({artist: artist}, {success: function (data) {
-                for (var i = 0; i < data.artist.image.length; i++) {
-                    if (data.artist.image[i].size === size) {
-                        $(images).attr('src', data.artist.image[i]['#text'] || defUrl)
-                    }
+
+            var _this = this
+            mopidy.library.lookup({'uris': [uri]}).then(function (resultDict) {
```
kingosticks: Shouldn't this try a `mopidy.library.getImages` first? Am I missing something?

mopidy_musicbox_webclient/static/js/images.js
```diff
@@ -61,47 +64,92 @@
...
+
+                var params = {}
+                if (artist.id) {
+                    params.id = artist.id
+                }
+                if (artist.name) {
+                    params.artist = artist.name
+                }
+
+                // TODO: Artist name may not be unique, which could result in the wrong images being retrieved.
```
kingosticks: Turns out this is exactly the issue I was having. Since we don't use Mopidy's `getImages` it falls back to last.fm and then picks the cover art for the wrong artist.

mopidy_musicbox_webclient/static/js/process_ws.js
```diff
@@ -229,7 +229,7 @@ function processArtistResults (resultArr) {
     resultsToTables(resultArr, ARTIST_TABLE, resultArr.uri)
     var artistname = getArtist(resultArr)
     $('#h_artistname, #artistpopupname').html(artistname)
-    getArtistImage(artistname, '#artistviewimage, #artistpopupimage', 'extralarge')
+    coverArt.getArtistImage(resultArr[0].uri, '#artistviewimage, #artistpopupimage', 'extralarge', mopidy)
```
kingosticks: I would have expected to pass the artist (rather than track) uri in here since we have it.

mopidy_musicbox_webclient/static/js/images.js
jcass77: This is easy enough to change and I agree using track artist before album artist will probably make sense for most situations.

mopidy_musicbox_webclient/static/js/images.js
jcass77: I have it as `mopidy.library.getImages` just retrieving **album** cover art. We're looking for the generic **artist** image here and need the track as a starting point.

mopidy_musicbox_webclient/static/js/images.js
jcass77: Yes. I'm not sure what to do about this except insist on using unique `MusicBrainz IDs` (which I think will not be available in most instances).

We can probably retrieve the album info from last.fm first, and then rely on the artist info that it sends back to make sure that we retrieve the correct image. It might require a bit of rework on the UI side, but I'll take a quick stab at it.

mopidy_musicbox_webclient/static/js/process_ws.js
jcass77: This was an oversight on my part to start with, but we'll probably end up doing a lookup on the track at this point anyway so that we can ask last.fm for the track info first, and then retrieve the artist image using the artist's MusicBrainz ID that it provides.

mopidy_musicbox_webclient/static/js/images.js
kingosticks: Sure, I appreciate the different goals here. My point is that if you pass getImages an artist_uri then, depending on the backend, it may be able to provide you with an artist image. If not, you can fallback to using track metadata to do a last.fm request.

jcass77: Thanks Nick, I didn't realise that backends are supposed to implement `getImages` to cater for artist URIs as well (don't think I've ever seen that happen before).

It makes sense to try that first and fall back to the track metadata after.

mopidy_musicbox_webclient/static/js/images.js
kingosticks: How would you do that? Do a last.fm artist search (I assume you can, not checked) first and then try and pick the correct artist from multiple results, at which point you'll then have the musicbrainz Id to use for image requests? Would the UI be to help pick the correct artist or to just flag the image as wrong and try again? Both sound like a fair bit of work.

mopidy_musicbox_webclient/static/js/images.js
kingosticks: Mopidy-spotify should, it can get images for any resource I think. Mopidy-tunein handles either artists or albums (can't remember which, it's a bit arbitrary) as station and show logos. Others, not so sure.

jcass77:
02cdb90ee8b7be3c9a7d55a09b832878c7f04936 is pretty much a complete rewrite, and addresses all of the comments.

The PR now also ensures that artist images will only be retrieved from last.fm using the unique MusicBrainz identifier (avoids having the wrong image being displayed if multiple artists with the same or similar names exist in the last.fm database).

jcass77:
Any further comments / suggestions for this PR before we merge?

kingosticks:
It's working great but I've not had a chance to peruse the code quite yet, Will do so at lunchtime.

mopidy_musicbox_webclient/static/js/gui.js
```diff
@@ -128,7 +128,7 @@ function setSongInfo (data) {
     }
     if (data.track.album && data.track.album.name) {
         $('#modalalbum').html('<a href="#" onclick="return library.showAlbum(\'' + data.track.album.uri + '\');">' + data.track.album.name + '</a>')
-        coverArt.getCover(data.track.uri, '#infocover, #controlspopupimage', 'extralarge', mopidy)
+        images.setAlbumImage(data.track.uri, '#infocover, #controlspopupimage', mopidy)
```
kingosticks: We could move this out of the if statement and leave `setAlbumImage` to sort `'images/default_cover.png'` for us.

mopidy_musicbox_webclient/static/js/process_ws.js
```diff
@@ -229,7 +229,8 @@ function processArtistResults (resultArr) {
     resultsToTables(resultArr, ARTIST_TABLE, resultArr.uri)
     var artistname = getArtist(resultArr)
     $('#h_artistname, #artistpopupname').html(artistname)
-    getArtistImage(artistname, '#artistviewimage, #artistpopupimage', 'extralarge')
+    var artistUri = resultArr[0].artists[0] ? resultArr[0].artists[0].uri : ''
```
kingosticks: Isn't `resultArr.uri` always the uri you want?

mopidy_musicbox_webclient/static/js/images.js
```diff
@@ -4,105 +4,279 @@
...
+
+            return {mbid: musicBrainzID, name: artistName}
+        },
+
+        /* Utility function for retrieving artist informaton for the given track from last.fm */
+        _getLastFmArtistInfo: function (track) {
+            var artist = images._getArtistInfo(track)
+            var artistPromise = $.Deferred()
+
+            if (!track || (!track.musicbrainz_id && !track.name) || (!track.musicbrainz_id && (!artist || !artist.name))) {
```
kingosticks: Isn't the mbid enough even without `track.name` or `artist.name` ?

mopidy_musicbox_webclient/static/js/images.js
```diff
@@ -4,105 +4,279 @@
...
 
-        getArtistImage: function (artist, images, size) {
-            var defUrl = 'images/user_24x32.png'
-            $(images).attr('src', defUrl)
-            if (!artist || artist.length === 0) {
+        /* Set artist image using the supplied artist MusicBrainz ID. */
+        _setLastFmArtistImage: function (mbid, img_element, size) {
+            if (!mbid) {
+                // Avoid expensive last.fm call if tag information is missing.
+                $(img_element).attr('src', images.DEFAULT_ALBUM_URL)
```
kingosticks: Should this be `DEFAULT_ARTIST_URL` ?

kingosticks:
The fact that this has tests is actually mind blowing. Really great work.

mopidy_musicbox_webclient/static/js/gui.js
jcass77: Yes. Good catch!

mopidy_musicbox_webclient/static/js/images.js
jcass77: Yes. Isn't that what this if testing for? It is confusing because it is stated in the negative, but I read it as:

1. we need a track, and then
2. we need either the mbid, or the track name
3. if we do not have mbid, then we need the artist name as well

So we should then be able to do a lookup using either ``mbid``, or ``track.name`` **AND** ``artist.name``.

mopidy_musicbox_webclient/static/js/images.js
jcass77: Absolutely. Isn't it impressive how I managed to get this mixed up, and then write a test that passes as well?

mopidy_musicbox_webclient/static/js/process_ws.js
jcass77: You're right, I'll change it.

jcass77:
Thanks for reviewing. I've addressed all of the comments in 3a7539aa02ce4eb5e2a47ef4460698e7a1f0cebe.

mopidy_musicbox_webclient/static/js/images.js
kingosticks: Yeh you are right but just writing those conditions might be clearer:
```
if (! (track && (track.musicbrainz_id or (track.name and artist and artist.name))))
```

jcass77: Done. I also applied the same structure change to line 80.

kingosticks:
LGTM. Squash and merge (that is a pretty handy new feature)?

Issue #184, opened 2016-03-16
Optimise 'updatePlayIcons' function


jcass77:
The ``updatePlayIcons `` function (https://github.com/pimusicbox/mopidy-musicbox-webclient/blob/88ff29db4c30bd4abfb248efad994b3da3123009/mopidy_musicbox_webclient/static/js/gui.js#L623-L649) does not scale well for very large tracklists.

It is responsible for inserting the 'playing' icon next to items in the various track lists to indicate which track is currently playing. The problem is that it loops over every item in the list every time the track changes, and this is made worse by the fact that we could have the track appearing in multiple lists (e.g. on the Queue, Playlists, Browse, and Search panes simultaneously). The performance degradation becomes noticeable if the tracklist contains more than a few hundred tracks.

I'm not quite sure how best to optimise this in a way that will ensure that MMW always stays in sync, especially considering that we may miss a Mopidy playback state change event every so often (e.g. if the connection to Mopidy is broken for whatever reason).

PR #193, opened 2016-04-24
Tracklist popup menu overhaul

### LINKED ISSUE #124: Unplayable tracks in a browse listing cause the incorrect track to be selected
Noticed this with SoundCloud, but if there are unplayable tracks (some tracks appear in the liked list but turn out to be 404 errors when you try and use their URIs) in the browse view, then they will be removed from the tracklist in Mopidy. This means that the client side code to play the selected track from the list selects the wrong track as the index doesn't get updated.

The code I changed uses track URI to select the track, rather than incrementing thru a tracklist.

I submitted my change here: https://github.com/colinnuk/mopidy-musicbox-webclient/commit/2eb4141f72ce1ec2ba60b5953ab150d2c9c7b268

But I just realised your develop branch has a rather different function entirely, which could fix the issue?

### LINKED ISSUE #133: Let users choose default click action
As discussed [here](https://github.com/pimusicbox/mopidy-musicbox-webclient/pull/36), the default action of replacing the whole queue with all tracks is not universally preferred.

I would like to see an option (either saved in the browser, or - preferably - set on the "server", in a line of code, maybe) to have the default click action be "Add to bottom of queue".

I tried to apply the pull-request myself, but the codebase has changed too much, apparently, and I am not familiar enough with it, to know where changes have to be made.

jcass77:
I've reworked the popup menus and refactored a few other small things along the way:

skitched-20160424-132044

- Changed the look and feel of the popups.
- Modularise ``controls.js`` so that it can be tested.
- Fixes #124: now uses ``tlid`` exclusively for playback, so playing the wrong track should no longer be an issue. Unplayable tracks will just follow the current Mopidy skip-to-next convention.
- Fixes #133 (sort of): now remembers the last popup action that was selected, and makes that the new default `onClick` action. The default is stored in a cookie to accommodate the preferences of different users using the same Mopidy backend. I'm not sure if that will work for everybody though. Ideally this should be a configurable default, but it will have to wait until the 'Settings' page has been enhanced to accommodate custom settings for MMW in addition to the Pi MusicBox stuff as well. Alternatively we could also specify that the user should perform at least two consecutive actions before the default is changed. I'm open to feedback / comments / suggestions on this.
- A lot of code refactoring and consolidation behind the scenes. 
- Tests for the main functions that I worked on, but there is still a long way to go.

There are some follow-up actions that I would like to submit in a separate PR once this has been merged:
- The popup functionality should ideally be split out into separate .js and .html files to make the code structure more modular.
- There are a few global arrays that are being used to "cache" tracklist information (see ``popupData``, ``playlists``, and ``customTracklists`` in ``functionfars.js``. The contents of these are difficult to predict at any given point in time, as various parts of the code will overwrite values, and some store ``mopidy.models.Refs`` while others store ``mopidy.models.Track``. It's probably time to refactor this into a proper centralised cache that can be used to route ``mopidy.library.lookup`` calls through.


jcass77:
It took a few attempts, but 53d02cf4d1b14363e2ea008086ad2ff23e8a5bef fixes the last of the tox build errors - mostly as a result of new default rules introduced in ``eslint-config-standard@5.2.0``, some timing issues in initialising the jQuery Mobile HTML stubs for testing, and my misunderstanding of how to clean up ``sinon.spy`` objects that are created over custom APIs.

kingosticks:
I'm not sure about having the default action dynamically change like this. User interface needs to be predictable and something stateful like this would not be. Particularly if multiple people are using the same shared client. Normally I'm against extra settings but this one has been requested many times (and only as a global seeing, not per user IIRC).

kingosticks:
In addition to these changes what do you think about tweaking some of the popup wording for consistency and brevity? I.e.
* "Add Track to Bottom of Queue" -> "Queue Track Last"
* "Play Track Next"  ->" Queue Track Next"
Etc

kingosticks:
And maybe icons for the last two menu options? A portrait/headhot one for artists and a CD for album? I've not looked what's available

jcass77:
I'll take a look at adding a configuration setting for this. If we're going to the trouble anyway then perhaps we can just as well add a ``DYNAMIC`` or ``USE_LAST`` configuration option which would use the logic originally proposed in this PR.

Changing the wording is easy enough. I've always thought that the plain-english approach is the easiest to understand. And seeing as we do not have any user documentation having overly verbose descriptions are probably ok.

We **could** add more icons for the 'Show...' options, but that would just be adding eye candy. The idea with the other icons is that they are functional: either showing what the current default ``onClick`` action is, or aligning with other icons on-screen to give the user more information as to what the option does (e.g. the 'cross' icon is used in the popup and on the queue page to indicate that it removes items from the queue).

I'm not against adding more icons if it does not end up just cluttering the user interface.

kingosticks:
> If we're going to the trouble anyway then perhaps we can just as well add a DYNAMIC or USE_LAST configuration option which would use the logic originally proposed in this PR.

I think we just need to be careful of creating a feature just because we can vs having a feature people want and will use. User interfaces that change are frustrating and take time for users to understand. I appreciate the logic is already in the PR and throwing away code is annoying but I am not convinced about this extra complexity.

> We could add more icons for the 'Show...' options, but that would just be adding eye candy. The idea with the other icons is that they are functional: either showing what the current default onClick action is, or aligning with other icons on-screen to give the user more information as to what the option does (e.g. the 'cross' icon is used in the popup and on the queue page to indicate that it removes items from the queue).

To be clear, I'm only talking about the popup menu icons you've got in the picture above. I think it looks odd for those two bottom menu options to have no graphic. I've not pulled and tested the changes so maybe I'm missing the functional aspect you are describing. 

I've also not gone through much of the diff as it is pretty unpleasant when viewed on github.

jcass77:
Thank you for the review @kingosticks .

I've added the ``on_track_click`` configuration option and squashed all of the commits.

Once you've given it a test run we can take a another look at adding the icons for the 'Show...' menu options if you still feel that it would make the interface easier to use.

The ``DYNAMIC`` functionality is encapsulated fairly neatly in https://github.com/pimusicbox/mopidy-musicbox-webclient/blob/581734aedf38e01d1570f89646a967a8756afc39/mopidy_musicbox_webclient/static/js/controls.js#L90-L109, so I don't think it is too bad.

jcass77:
And, apologies for the large diff. It's mostly just indentation changes and adding function prefixes as a result of wrapping ``controls.js``.

kingosticks:
Having all the icons change whenever I select a different option from last time (wasn't there supposed to be hysteresis?) is a bit weird. All the extra icons make it visually quite busy and they are only necessary when you're in the dynamic mode (I do see that without some visual feedback that mode is unusable).

jcass77:
Pushed e9726a30081a31e795aab7fc2d406b78c4c640fa to cater for Mopidy < 2.0.0 in which ``tlid`` can be ``0``.

jcass77:
I don't think the icons should update when you select a different option unless you are in dynamic mode - this is an oversight on my part which I'll fix tomorrow.

It does look slightly busier compared to just having the ellipses. I feel the extra feedback to the user on what the current default is is handy but this is also something that we can just restrict to the dynamic mode if that would be preferable.

jcass77:
cb5ceafab79ffcb1eabda11b88436034de301e6c addresses the last few comments - I think it is ready for review.

jcass77:
7c2e38e623b2e6e94cf917fa20688f8fa3ade313 fixes an issue where the 'now playing' icon would sometimes not appear next to the currently playing track in the list.

jcass77:
I think I am going to merge this PR before I am tempted to bloat it even further. 

I've implemented pretty much all of the optimisations that I can think of for #184 as well, mainly by just reducing the number of DOM manipulations as much as possible by checking which classes are assigned to the element before changing it.

Please let me know if there are any other comments or suggestions and we can circle back.

kingosticks:
You know you can have multiple branches and rebase your other branches on develop after this is merged, right? 

I've not had enough time to go through the diff as it's big and it affects the playtrack functions which are historically difficult to get right. One thing I noticed was that you did seem to remove the special PLAY_ALL case for a stream. This one might seem odd but it's actually surprisingly useful behaviour if you try and use it. Am I right this has now gone?

jcass77:
Yes I thought about adding more branches - may I'll follow that approach next time instead.

You mean the bit where streams would always ``PLAY_NOW`` instead of ``PLAY_ALL``? Now that streams are just treated as regular tracks in the 'Playlists' pane I did not think that was necessary anymore. And ``playStreamUri`` seems to clear the queue before starting playback anyway.

I might have missed something though - could you describe the situation in which the old functionality was more useful?

kingosticks:
> You mean the bit where streams would always PLAY_NOW instead of PLAY_ALL? Now that streams are just treated as regular tracks in the 'Playlists' pane I did not think that was necessary anymore.

Yes. But no, it's not to do with that. It's the case where you click a particular stream from a long list of results. Having all those streams added to your queue is almost never what you actually want to do.  And in the case of unreliable internet streams, if the one you clicked doesn't work then Mopidy will then spend time trying all the streams in turn until it finds one that works. Any old subsequent stream can then start playing and it's really quite confusing what happened. It's probably worth trying this for yourself to see what I mean. If we had some better way to feedback the fact the original stream failed to play then perhaps it would be less confusing.

jcass77:
Oh I see. It's a little strange that streams should be handled differently though. It makes even less sense now that the default click action is user-configurable - does that imply that we should cater for defining a different onClick default for each Mopidy scheme (especially seeing as there may be a few backend extensions that serve up 'streams')?

Is just selecting 'play now' from the popup menu not an acceptable course of action if the user wants to avoid having everything added to the queue?

kingosticks:
I agree it seems strange and I also got rid of it at one point. But then when I started using it (rather than just testing it) I got frustrated and realised it was actually quite helpful. 

Streams are a slightly special case due to the fact they can be so very unreliable.

But you are right, it does make less sense now the default click action is configurable.. I don't think we need to care if it makes the code weird **if** it substantially improves the user experience and Just Works. I'm not sure what is best any more.

kingosticks:
Hmm let's just keep this how it is for now and we can revisit it post-release if need be.

kingosticks:
I'm really not a fan of the skip icon being used for "play all", it's just weird to overload this here. Something like [this](http://plainicon.com/download-icon/47732/play-all) makes more sense assuming it's still meaningful when scaled down to the size we need to use.

And in general, what do you think about reverting to the normal 3 dots popup icon unless in dynamic mode? The inconsistency with the queue popup icon is a little odd looking.

jcass77:
> I'm really not a fan of the skip icon being used for "play all", it's just weird to overload this here. Something like [this](http://plainicon.com/download-icon/47732/play-all) makes more sense assuming it's still meaningful when scaled down to the size we need to use.

I like the alternative icon that is being proposed but it would be **so** much simpler if we could stick to using the FontAwesome library rather than adding another icon to the repository (which we always seem to lose track of between deleting unused icons and keeping mb.appcache in sync). The icons are also currently updated by just changing the ``class`` attribute, so adding an ``img`` element in there will be a special case that we need to cater for.

Our options are basically to use one of the other "play icons" (e.g. https://fortawesome.github.io/Font-Awesome/icon/play-circle or https://fortawesome.github.io/Font-Awesome/icon/play-circle-o)

...or create a custom stack with something like:

```html
<span class="fa-stack fa-lg">
  <i class="fa fa-clone-o fa-stack-2x"></i>
  <i class="fa fa-play fa-stack-1x"></i>
</span>
```
The latter really complicates updating the icons in dynamic mode and is probably more effort than it is worth.

> And in general, what do you think about reverting to the normal 3 dots popup icon unless in dynamic mode? The inconsistency with the queue popup icon is a little odd looking.

I don't feel strongly about this either way. After using it for a while I found it comforting to always have an on-screen reminder of what ``onClick`` will do (especially having multiple MusicBoxes running in the house that may be configured differently, and also doing development on my laptop which is constantly being reconfigured for testing). I'll defer to you on this one so let me know if you want it changed.

kingosticks:
I do agree, FontAwesome is both awesome and yet frustrating at the same time. `play-circle-o` probably gets my vote. 

I'm not convinced users would intentionally configure their musicboxes differently, that seems like a recipe for a headache. The dots seem much kinder to my eyes but I think that's partly because they stand out less compared to the contrasting black on grey you get with the play icon. The popup button shouldn't be jumping out at you, more of a background, secondary concern (IMO).

PR #195, opened 2016-04-27
fix:Replace JavaScript prompt with jQuery equivalent.

### LINKED ISSUE #191: Replace Javascript confirmation prompt
A JavaScript confirmation prompt is displayed when the user tries to delete a stream: https://github.com/pimusicbox/mopidy-musicbox-webclient/blob/f43a9a7afaffc50e3abe278a3df36203f8ca7308/mopidy_musicbox_webclient/static/js/controls.js#L577

This should probably be replaced with a proper jQuery Mobile popup instead to ensure a more consistent look and feel throughout the application.

jcass77:
Fixes #191.

Would have been be nice to avoid the second call to ``controls.getFavourites()`` - but that probably requires refactoring other areas of the code as well.


jcass77:
0b740adac9444685209950680cc2ecf9eaed6b05 centre-aligns the text in all popup dialogs.

jcass77:
1b75257682571d67e30f2bfa1c96301da5f5ecbc tries to standardise the popup dialog layouts a little:

- sentence fragments should not have any punctuation.

- buttons that perform a destructive action go on the left

- make the buttons more descriptive to re-affirm the action that they will perform.

Loosely based on [these guidelines](https://developer.apple.com/library/ios/documentation/UserExperience/Conceptual/MobileHIG/Alerts.html) or, if you are not an Apple fanperson and habitually take issue with whatever they recommend, [these](http://uxmovement.com/buttons/why-ok-buttons-in-dialog-boxes-work-best-on-the-right/).

jcass77:
I think this works. If there are no further comments I'll go ahead and merge...

PR #197, opened 2016-05-04
fix:Small divider should not be rendered before last track in an album.

### LINKED ISSUE #196: Last album track not grouped with the rest in results list
The last track in an album doesn't seem to be correctly grouped with the rest of the album's tracks and gets a wider divider, as if the track is from a different album. The screenshot below is from a Spotify Artist view so the metadata should be correct. The behaviour is the same across all views (since it all uses the same code, I think!)

last-album-track

jcass77:
Fixes #196.


kingosticks:
Fantastic, thanks.

PR #198, opened 2016-05-05
Alternative icon set

jcass77:
This PR proposes an alternative set of tracklist action icons, which forms part of the discussion sparked by #193.

- Make the UI more consistent: action options for individual tracks will be displayed above those that can be performed on *all* tracks. Similarly, action icons for individual tracks won't have the background set, while icons that effect all tracks will have a filled background.

- Make the popup menu icons light grey so that they are less obtrusive.

- Extend the convention of having the popup menu icon always represent the default ``onClick`` action to the Queue pane.


kingosticks:
What you think about making the popup icon a bit larger? It's tricky to make out the different play ones.

When using `DYNAMIC` mode Is the queue supposed to remember it's last action?

kingosticks:
And I like what you've done with the consistent approach in your first point.

jcass77:
> What you think about making the popup icon a bit larger? It's tricky to make out the different play ones.

I named one of the CSS properties incorrectly. It's fixed in 2ba008864f8bd030d3fb81ac995b932f18d5c4be which should increase the icon size slightly.

> When using DYNAMIC mode Is the queue supposed to remember it's last action?

Not at this stage since the only other option is to remove the track from the Queue. I didn't think that it would be very useful to set the default `onClick` behaviour to have the track disappear from the queue so did not spend the time on it. It also means having two separate configuration settings as the actions that can be performed in the various popups do not overlap perfectly.

But we can revisit this if anybody asks for such a feature.

jcass77:
...just realised that I was also using the wrong 'PLAY_ALL' icon for the popup menu in the Queue pane. It should show the 'PLAY_NOW' icon instead - fixed in eac5c6412276f6c6963458c6b58936dd9d54e530

kingosticks:
> I named one of the CSS properties incorrectly. It's fixed in 2ba0088 which should increase the icon size slightly.

That explains why my attempts to play around with the size were so unsuccessful!

kingosticks:
Trying to set `large` for the size creates an interesting pattern effect but it's not what we are looking for. Are they visible enough on a small screen (i.e. smartphone)? I cannot test it on my phone while at work. If so then I think all my concerns are now resolved.

kingosticks:
Ah, except we should probably add the new config option to the README since it's something that most people might want to play with.

jcass77:
I've updated the README. @kingosticks, take a look and perhaps fine-tune further when you update the README in preparation for the next release?

I've also included a new screenshot since the current one is quite old, and some of the user interface elements have changed. It will only render properly on github once this PR lands in the develop branch though.

kingosticks:
Looks good to me on a small screen and thanks for adding that extra README stuff.

PR #202, opened 2016-05-14
Avoid album cover lookup when clearing the queue

### LINKED ISSUE #201: Clearing the queue triggers a Application error
When we clear the queue we call `resetSong()` to clear the current song info by [manually populating a fake Mopidy Tltrack object](https://github.com/pimusicbox/mopidy-musicbox-webclient/blob/64eda54a78ebe90c31fdc8003a86533ae0108632/mopidy_musicbox_webclient/static/js/gui.js#L12-L19) and passing it to `setSongInfo()`. We then end up trying to get images for the uri `' '` (a space character) which Mopidy doesn't like. Mopidy log available [here](http://pastebin.com/rG0LmB2n).

```
{"message": "Application error", "code": 0, "data": {"message": "Expected a list of URIs, not u\' \'",
```
This is kind of related to https://github.com/pimusicbox/mopidy-musicbox-webclient/issues/172

jcass77:
Fixes #201.

We actually already check if the URI is valid with ``if (!uri)`` to avoid exactly this problem, but since the song URI is reset to ``' '`` (space) the check fails. I can't think of a reason why it was done this way originally and a quick test did not reveal any immediate issues with just using ``''`` instead.

As for creating the fake ``TlTrack`` object - that is probably something that should be refactored out, but I am hesitant to do that just now since ``gui.js`` has not been modularized yet so any changes that we make cannot be tested. And ``setSongInfo()`` looks fairly brittle.


kingosticks:
Yeh, I did notice that. Perhaps there used to be some other guard for an empty string that this trick was trying to avoid, which is not a great idea 

But I agree, this fix is fine for now.

PR #215, opened 2016-12-04
Detect if MMW is running on 'MusicBox' and show correct server info.

### LINKED ISSUE #67: Show the IP-Adress of the musicbox-server in webclient
Hi,

from my point of view it would be very handy to make the current IP-Adress of the musicboxserver visible in the Webclient.

Regards

Chris

### LINKED ISSUE #209: Use correct hostname in "Trying to reach MusicBox" popup when Mopidy server is unreachable
If a websocket connection cannot be made, the popup message says "Trying to reach MusicBox, please wait". In the case of a non-musicbox install this confusing for the user. It should use the target system's hostname, or a more generic message.

jcass77:
Show server information in loader and in navigation pane (hostname, ip address, port).

skitched-20161204-122233

Fixes #67. Fixes #209.


jcass77:
@kingosticks, are you happy for me to merge this?

kingosticks:
Sorry John I've been a little busy this past week. Let me take a real quick look this morning. I was thinking that long hostnames might look pretty strange when wrapped (assuming they will wrap) in the confines of the thin side menu but I guess there's no alternative to that

kingosticks:
Do you know what info you get when Mopidy is running on localhost behind a proxy? That case (which musicbox has/will switch to) will surely see requests to localhost:6680 which isn't really helpful to the user. But again, not sure how to fix that.

kingosticks:
Turns out the localhost thing can be entirely solved in the proxy configuration. So that's a non issue.

jcass77:
>  I was thinking that long hostnames might look pretty strange when wrapped (assuming they will wrap) in the confines of the thin side menu but I guess there's no alternative to that...

I've pushed c264bc58b9619b7d389bbc1d7265169bbc3d7078 which wraps better and adds ellipsis for long strings, so the worst it can get is:

skitched-20161211-063612

PR #219, opened 2017-01-06
More appropriate icons when browsing local folders.

jcass77:
When browsing the local folder structure, the 'folder' icon is used for all folders, audio files, and other files.

This PR enhances MMW to use more appropriate icons for each, which should make browsing a little easier:

skitched-20170106-170202


kingosticks:
Awesome. LGTM

PR #220, opened 2017-01-08
Music upload feature

fbarresi:
I added a new handler and a new page for music upload.
Of course I needed to add a new optional setting in the configuration in order to let the user decide the target path of the uploaded music.
The user can add as many files as he want at the same upload and specify a subfolder.
The music uploaded can be immediately played if the upload path is even in the mopidy local media settings.
I done that thinking about a growing music library that should be easy extended with new tracks/albums or about a sort of "party mode" for Musicbox where people share music together.


mopidy_musicbox_webclient/static/upload.html
```diff
@@ -0,0 +1,97 @@
...
+    </div>
+
+<div data-role="content" class="pane" id="uploadpane">
+  <div class="ui-grid-a pl-breakpoint">
+{% if can_upload %}
+      <h4>Uploading into: {{upload_path}}</h4>
+      <div class="ui-block-a" style="padding: 5px">
+        <form enctype="multipart/form-data" method="post">
+          <p>
+            Subpath: <input type="text" name="subpath" />
```
jcass77: This might be a little input-error-prone as user's won't be able to see which subfolders already exist from this page, and typos will happen.

kingosticks:
I'm not sure where this best belongs, I'd probably lean towards it being a separate mopidy extension as its not really anything to do with any particular web client. But I'm open to it if there's a particular reason. But since we have to maintain it then it needs a good reason. 

In a 'party' use case you probably won't get anyone digging around context menus. And nobody will take the time to respect your folder structure. You want a big obvious button to play your music files now. 

If the use case is the admin/single user remotely updating their music collection then I think that's a job better left to dedicated tools (ftp, shares) with full support for renaming etc. 

If it's integrated then it does need a way to be disabled (and hidden?). Providing users with write access to the server is inappropriate for some musicbox installations. 

I agree with @jcass77 on the rest, particularly with regards to maintaining a separate page.

jcass77:
There's been a resurgence in Mopidy extensions specifically developed for hosting parties (searching for 'mopidy party' or 'mopidy juke' on GitHub will yield four or five results, many of them recently updated). All of those extensions will probably be more suited and work better for those types of use cases, so if that is what we are aiming for then I don't think this feature belongs in MMW.

I guess I am not sure how often this will be used as it is not a feature that I have ever seen requested in any of the forums or the repositories of the other extensions. If we can get all of the suggestions to this PR addressed then it might be convenient for uploading the odd file to a specific directory every now and again.

A few test cases might also make this easier for future contributors to maintain and add to.

fbarresi:
Thank you for your impressions, very useful!
This pull request never wanted to be something like: "hey there, look at this new feature that I draft. Now YOU can maintain and improve it". 
I just proposed this change because I think, as Musicbox fan, that it is useful. I also whould be glad to collaborate in the project making this draft a better integrated part of the Webclient.
Meanwhile I think you are right and it should be better an indipendent Extension of Mopidy.
I "declare" this pull request closed.

kingosticks:
Thanks for taking the time to prepare the PR, we don't get many people who are interested in / capable of helping so it is appreciated. Please do consider talking to some of the 'party orientated' mopidy clients out there, or having it standalone. It's a nice idea, especially combined with the file extension (where you don't have to rescan the database before seeing the new uploaded files).

PR #221, opened 2017-01-14
Initialize GUI before attempting to connect to Mopidy

jcass77:
This addresses a long-standing issue where the UI would not initialize properly if the Mopidy server cannot be reached.

Will now always show the 'Trying to connect...' popup and hide the inactive panels.


jcass77:
After a few days of testing, I think this works well - merging.

PR #222, opened 2017-01-14
Add 'Refresh' button to update libraries

### LINKED ISSUE #136: Feature: Ability to refresh backend libraries
At the moment, there is no way (that I know of at least) to invoke `backend.LibraryProvider.refresh()` from within the UI.

This would be a nice feature to add to allow backends to be refreshed when the underlying library changes as many backends appear to be employing some sort of library caching mechanism.

Mopidy-Mobile implemented this using a `drag down` swipe, but we could also just re-use the refresh button approach that came out of https://github.com/pimusicbox/mopidy-musicbox-webclient/issues/107.

jcass77:
Add refresh button to 'Browse' pane.

Ensure that only playable music files have context menus.

Fixes #136.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -119,8 +119,6 @@ v2.4.0 (UNRELEASED)
   (Fixes: `#213 <https://github.com/pimusicbox/mopidy-musicbox-webclient/issues/213>`_).
 - Now shows correct hostname information in loader popup. (Fixes: `#209 <https://github.com/pimusicbox/mopidy-musicbox-webclient/issues/209>`_).
 - Reset 'Now Playing' info when the last track in the tracklist is deleted. Fixes an issue where info of the last song played would be displayed even after the queue had been cleared.
-- Now initializes the GUI properly, even if the user is offline or the Mopidy server cannot be reached.
-- Fixed `Alarm Clock <https://pypi.python.org/pypi/Mopidy-AlarmClock/>`_ detection.
 - Use correct icons for folders, audio, and other files when browsing local files.
 - Now initializes the GUI properly, even if the user is offline or the Mopidy server cannot be reached.
 - Fixed `Alarm Clock <https://pypi.python.org/pypi/Mopidy-AlarmClock/>`_ detection.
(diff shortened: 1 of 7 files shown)


jcass77:
I've rebased this on 'develop' and refactored the PR to make it smaller and easier to review.

It essentially does just two things:

- add the 'refresh' button when browsing the library, which just calls ``mopidy.LibraryController.refresh`` with the current folder's URI.

- ensure that only playable tracks will have an icon for opening the context menu, when browsing files.

I also played around with using different icons to represent folders, files, and playlists so that they are easy to recognize anywhere in the UI, but in the end decided that these looked a little tacky so rolled that back.

kingosticks:
Was it really the idea here to remove all the per-media type icons defined in uriClassList like we used to have? This was very useful, particularly for the playlists as it's the only UI hint as to their source.

jcass77:
Can you elaborate on where the icons are not rendering as expected for you?

Each service shows its own icon on the 'browse' pane. I can't recall whether tracks also used to use these service icons - the problem with that was that you couldn't distinguish tracks from folders. I might have changed it so that all playable tracks just show the 'sound' icon. Is that what you are reporting?

skitched-20170218-073800

Every level of the folder hierarchy also uses the service icon as you drill down. This helps to keep track of which service you are working with as some of the hierarchies can be fairly deep:

skitched-20170218-073900

But playable tracks show the sound icon. This lets you know that clicking on it will start playback, add it to the queue, etc:

skitched-20170218-074014

Queues and playlists show either the sound or stream icon:

skitched-20170218-074226

kingosticks:
I was talking about playlists, playlist entries and search results.

jcass77:
Those all show the 'sound' icon if it is a playable track, and when clicking on the 'Show Info' menu option the track URI will indicate which backend is it from.

Would you like to change any of this?

kingosticks:
I much prefer the old way where the backend was obvious without having to show info. Maybe unplayable tracks could be greyed out or something. I can look into it.

PR #223, opened 2017-01-15
Fix alarm clock detection

DavisNT:
Looks like AlarmClock detection has been broken during refactoring (in commit 411a3781fdab270984299e47f136abec8395ff57). 
This commit fixes alarm clock detection (and has been tested).

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -113,6 +113,7 @@ v2.4.0 (UNRELEASED)
 
 **Fixes**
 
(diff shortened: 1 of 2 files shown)


jcass77:
Thanks for the PR. I've updated the test case as part of 1c59940075bf170bf6e7ffa08d753f6058b0adf8 to ensure that we do not have a repeat of this issue.

DavisNT:
Great! Thanks a lot!

Issue #38, opened 2014-07-23
Allow scrolling playlists and playlist contents independently


kingosticks:
If you have more playlists than can be displayed and scroll down to select one at the bottom, you then need to scroll back up to the top to see it's contents. Ideally the list of playlists and the list of tracks in the selected playlist could be scrolled independently .

nickbarnwell:
@woutervanwijk are you still accepting contributions to Musicbox? I'd be happy to fix this and a few of the other outstanding issues+PRs this weekend if you're willing to CR.

woutervanwijk:
Sure, please contribute! It's more than welcome

kingosticks:
This is addressed as part of https://github.com/woutervanwijk/Mopidy-MusicBox-Webclient/pull/86

PR #224, opened 2017-01-22
Save scrolling position when switching pages or browsing the library.

### LINKED ISSUE #73: Some scrolling oddities when changing views
I did notice, along the lines of #71 and scrolling the view that the view isn't scrolled on view changes.

The scrollbar appears to remain at the same length value, which if the new view is shorter will result in arriving at the bottom of the page and if the new view was actually quite a bit longer it will be at some point in the middle. I think it would be nice to scroll to the top on new page view loads as a courtesy.

### LINKED ISSUE #93: browsing local files doesn't save scroll position (i have a fix for this)
I've got over 700 artists on a local drive and I want to scroll through them to add songs to my queue (i've got the feature to add files to queue from browse)...

The problem is when I scroll down to say "The Rolling Stones", add a file to my queue, and go Back to the list of artists, it starts at the beginning. To get back to or past "The Rolling Stones" I have to flip through hundreds of artists again.

I think it would be nice if it scrolled to or stayed at the last artist you were browsing.

If the bug/enhancement is accepted I can put in a fix

My fix:
In library.getBrowseDir(), I append :scrollY=xxxx to the browse stack item if I'm going into a directory or I get the scrollY position off the popped stack item if I'm backing out.
In process_ws.processBrowseDir(), I call window.scrollTo(0,scrollY) if scrollY > 0.
I also clean up the browse stack item to make sure we query mopidy with :scrollY=xxx in there

jcass77:
Fixes #73. Fixes #93.

I'm not sure whether it makes sense to add UI state (like storing the scrolling position) to the ``browseStack`` array. But that seems better than declaring yet another array just to persist the UI state, and then have to keep multiple structures in sync using URI identifiers?


jcass77:
Pushed 0e6e15e8e5af8b2d3e234daa1b1af9105925adb1 which fixes an issue with the scroll position not being loaded correctly for the 'browse' pane.

kingosticks:
I intend to have a proper look at this later when I get back home. I always
envisioned we'd try and change to use scrolling divs for these
tracklist/playlists and that would mitigate some of the issues we have. The
way the scrolling currently works is really suboptimal.

jcass77:
This PR only addresses the issues with the current scrolling mechanism.

Using scrolling divs might be better as a next step (especially for the 'Playlists' page - #38).

jcass77:
Are we happy to merge this as an interim measure until we take a closer look at #38?

PR #228, opened 2017-01-29
Show all available track info in the 'Show Track Info...' popup.

### LINKED ISSUE #227: Add remainder of track properties to the 'Show Track Info...' popup
Not all of the available track information currently shows up in the track info popup.

The popup can probably be expanded to include:

- [x] genre
- [x] disc_no
- [x] date
- [x] comment
- [x] musicbrainz_id
- [x] last_modified

jcass77:
...the popup can now also be activated by either clicking on the little 'info' icon on the top right corner of album image, or the track title in the 'Now Playing' pane.

skitched-20170129-112123

Fixes #227.


jcass77:
Would anyone like to comment on this before I merge later today?

kingosticks:
Apologies, this kept slipping my mind. Have you tried this on a mobile device? On my phone, if you trigger the popup the focus goes straight to the uri textbox which opens the onscreen keyboard, all of that doesn't fit on my screen so you then have to close the keyboard to read everything. It might just be an Android thing. Why is the uri box different? So you can easily copy and paste it?

README.rst
```diff
@@ -108,8 +108,8 @@ v2.4.0 (UNRELEASED)
 
 - Now shows server name/IP address and port number at the bottom of the navigation pane. (Addresses: `#67 <https://github.com/pimusicbox/mopidy-musicbox-webclient/issues/67>`_).
 - Add ability to insert a track anywhere in the current queue. (Addresses: `#75 <https://github.com/pimusicbox/mopidy-musicbox-webclient/issues/75>`_).
-- Add 'Show Track Info' popup which can be activated from any context menu. The popup includes the URI of the track,
-  which can be inserted into various lists elsewhere in the player.
+- Add 'Show Track Info' popup which can be activated from any context menu or by clicking the 'info' icon next to the album cover on the
+  'Now Playing' pane. The popup includes the URI of the track, which can be inserted into various lists elsewhere in the player.
```
kingosticks: or the track's title text.

jcass77:
The ``isMobile`` check prevents the keyboard from popping up on mobile devices. I tested it on iOS and it is *supposed* to work for Android as well.

Yes, the URI box is different in order to make text selection easier on desktop (where it will pre-select the URI as soon as the popup opens).

mopidy_musicbox_webclient/static/js/controls.js
```diff
@@ -338,74 +338,129 @@
...
+                    html += stringFromTemplate(rowTemplate, row)
                 }
+
+                if (track.disc_no) {
+                    row = {'label': 'Disc #', 'text': track.disc_no}
+                    html += stringFromTemplate(rowTemplate, row)
+                }
+
+                if (track.date) {
+                    row = {'label': 'Date', 'text': new Date(track.date)}
```
kingosticks: Could we format the date to be more friendly? It displays things like 'Wed Jan 01 1997 00:00:00 GMT+0000 (GMT)' for me.

mopidy_musicbox_webclient/static/js/functionsvars.js
```diff
@@ -204,12 +204,14 @@ function getAlbum (pl) {
 function artistsToString (artists, max) {
     var result = ''
     max = max || 3
-    for (var i = 0; i < artists.length && i < max; i++) {
-        if (artists[i].name) {
-            if (i > 0) {
-                result += ', '
+    if (artists && artists.length > 0) {
```
kingosticks: I'm sure we have code like this all over the place, doesn't JS have a .join() method?

mopidy_musicbox_webclient/static/js/gui.js
```diff
@@ -316,7 +320,8 @@ function initSocketevents () {
     })
 
     mopidy.on('event:streamTitleChanged', function (data) {
-        setSongTitle(data.title, true)
+        // Update all track info.
+        mopidy.playback.getCurrentTlTrack().then(processCurrenttrack, console.error)
```
kingosticks: Out of interest, is this a bug fix or is this to make the popup also change, or something?

kingosticks:
My user agent looks like it would match so I'm a bit confused. 
```
Mozilla/5.0 (Linux; Android 6.0; HUAWEI VNS-L31 Build/HUAWEIVNS-L31) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/55.0.2883.91 Mobile Safari/537.36
```

mopidy_musicbox_webclient/static/js/controls.js
jcass77: The easiest approach, which appears to be more or less standard and well supported, is to use ``localeString()``.

mopidy_musicbox_webclient/static/js/functionsvars.js
jcass77: Yes you can call ``join()`` on an array and specify the separator. I think we can also drop the ``max`` argument as it is not currently used anywhere. I'm hoping that the frontend HTML has been written in a way that will add ellipses if the text becomes too long.

jcass77: ...the real issue is that the array contains artist objects and not just artist names. So we can't use join without copying the array first, which seems worse than iterating over it as we do at the moment.

mopidy_musicbox_webclient/static/js/gui.js
jcass77: Not a bug fix, but I think it is safer to run through all of the ``processCurrenttrack`` logic when the stream title changes (which also updates the UI and all of the popup menus).

jcass77:
> My user agent looks like it would match so I'm a bit confused.

>Mozilla/5.0 (Linux; Android 6.0; HUAWEI VNS-L31 Build/HUAWEIVNS-L31) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/55.0.2883.91 Mobile Safari/537.36

Yeah I'm stumped. Are you sure that the browser cache has been cleared?

If you manage to find a regular expression that works for your user agent then we can just update it.

jcass77:
I've pushed 69329435d2617eca6dd02b958e63b4f853e42066 which addresses some of the comments.

Not sure if there is much to be done about the rest?

kingosticks:
Awesome, thanks.

PR #260, opened 2019-02-09
Adding version info to home page

fmarzocca:
I thought it was useful to have musicbox version info on web Home Page (right after the host informations).

### DIFF
--- a/mopidy_musicbox_webclient/static/css/webclient.css
+++ b/mopidy_musicbox_webclient/static/css/webclient.css
@@ -632,6 +632,11 @@ a {
     color: #efefef;
 }
 
+#version-label {
+    position: fixed;
+    bottom: 0;
+}
+
 /*desktop*/
 @media (min-width: 55em) {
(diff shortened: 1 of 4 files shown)


fmarzocca:
Changes has been applied to move the version label at the bottom.

kingosticks:
Please remember that Mopidy-Pimusicbox-Webclient is a generic Mopidy webclient that can be installed on any system. It is also used in PiMusicbox. I still think this information is only interesting for debug, you don't need to see it all the time. It seems weird to have a generic Mopidy webclient display the *system* version. I assumed you were going to display the webclient version.

kingosticks:
And what happens on those other systems where /etc/issue doesn't exist?

fmarzocca:
>  I assumed you were going to display the webclient version.

where is the webclient version?

kingosticks:
You can get it from https://github.com/pimusicbox/mopidy-musicbox-webclient/blob/develop/mopidy_musicbox_webclient/webclient.py#L20

Issue #265, opened 2019-10-30
IPv6 is not supported


xiongyihui:
When a local domain such as `raspberrypi.local` is resolved to an IPv6 address, there is an unhandled error.

```
ERROR    Uncaught exception
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/dist-packages/tornado/http1connection.py", line 238, in _read_message
    delegate.finish()
  File "/usr/local/lib/python2.7/dist-packages/tornado/routing.py", line 256, in finish
    self.delegate.finish()
  File "/usr/local/lib/python2.7/dist-packages/tornado/web.py", line 2195, in finish
    self.execute()
  File "/usr/local/lib/python2.7/dist-packages/tornado/web.py", line 2215, in execute
    **self.handler_kwargs)
  File "/usr/local/lib/python2.7/dist-packages/tornado/web.py", line 194, in __init__
    self.initialize(**kwargs)
  File "/usr/local/lib/python2.7/dist-packages/mopidy_musicbox_webclient/web.py", line 52, in initialize
    'serverIP': socket.gethostbyname(url.hostname),
gaierror: [Errno -2] Name or service not known
```

The reason is that `socket.gethostbyname` [doesn't support IPv6](https://docs.python.org/2/library/socket.html#socket.gethostbyname).   `socket.getaddrinfo()` should be used instead

Code at https://github.com/pimusicbox/mopidy-musicbox-webclient/blob/25e29b4ed1/mopidy_musicbox_webclient/web.py#L52

kingosticks:
Fixed in #266

PR #266, opened 2019-10-30
support IPv6 address

xiongyihui:
As [`socket.gethostbyname()` only supports IPv4](https://docs.python.org/2/library/socket.html#socket.gethostbyname), use `socket.getaddrinfo()` instead to support IPv6 address.

Solve the issue #265


kingosticks:
Thanks!

PR #272, opened 2020-04-23
Support for the chrome Key Socket Media Plugin

faulker:
Plugin to support the Key Socket Media Keys chrome plugin. The plugin allows for controlling the music using the media keys.

https://github.com/borismus/keysocket
https://smus.com/chrome-media-keys-revisited/
https://chrome.google.com/webstore/detail/key-socket-media-keys/fphfgdknbpakeedbaenojjdcdoajihik?hl=en


kingosticks:
I don't use chrome myself but this looks cool. Surprised there's nothing cross-browser for this. Can I check none of this code requires any extra license?

faulker:
It shouldn't require any licensing, it isn't using any code from any other package it is just implementing listeners for triggered custom events that the plugin fires off.

Also if you don't like the auto plugin loading code let me know, I kind of implemented it for ease if I do any other "plugins"

kingosticks:
Super, thanks.

PR #279, opened 2021-02-13
Added the stop button

shameerariff:
Just added the stop button, probably it need testing as I do not have all the services installed in my env.

### DIFF
--- a/mopidy_musicbox_webclient/static/js/controls.js
+++ b/mopidy_musicbox_webclient/static/js/controls.js
@@ -528,9 +528,9 @@
         },
         
         doStop: function () {
-		    toast('Stopping current track...')
-		    mopidy.playback.stop()
-	    },
+                toast('Stopping current track...')
+                mopidy.playback.stop()
+        },
         
         doPrevious: function () {
             toast('Playing previous track...')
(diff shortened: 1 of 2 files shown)


mopidy_musicbox_webclient/static/js/controls.js
```diff
@@ -526,7 +526,12 @@
             }
             controls.setPlayState(!play)
         },
-
+        
+        doStop: function () {
+		    toast('Stopping current track...')
+		    mopidy.playback.stop()
+	    },
```
djmattyg007: These lines mix hard tabs with soft tabs. It appears that the rest of the file uses soft tabs. Would you be able to convert these to soft tabs?

shameerariff (commented):
Hard tabs are replaced with soft tabs

mopidy_musicbox_webclient/static/js/controls.js
```diff
@@ -526,7 +526,12 @@
             }
             controls.setPlayState(!play)
         },
-
+        
+        doStop: function () {
+                toast('Stopping current track...')
+                mopidy.playback.stop()
```
djmattyg007, line 532: The indentation on this line doesn't match the indentation in other functions on either side of this function.

djmattyg007 (changes requested):

